PR #8, opened 2022-03-02
Astractl 15308/install exe system calls

jharrod:
Adds install.go. Golang based install executable. 

* Loads image tar (optional)
* Re-tags and push images to private repo (if tar is loaded)
* Configures Astra Connector yaml
* Applys the Connector Operator yaml
* Applys the Connector yaml

Test options are there (skipTLS, astraURL, etc)


.gitignore
```diff
@@ -17,6 +17,7 @@ testbin/*
 # Kubernetes Generated files - skip generated files, except for vendored files
 
 !vendor/**/zz_generated.*
+deployConfig.yaml
```
jharrod, line 20: The installer writes to this file

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+		if err != nil {
+			return nil, err
+		}
+		imagesWithRepo = append(imagesWithRepo, newTag)
+	}
+	return imagesWithRepo, nil
+}
+
+////todo: fix auth issue here so we can stop pushing via system call
+//func pushImages(dockerClient *client.Client, opts *Options, images []string) error {
```
jharrod, line 158: See comment ^. 
Commented out for now to avoid unnecessary dependencies

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+	Stream      string       `json:"stream"`
+	Error       string       `json:"error"`
+	ErrorDetail *ErrorDetail `json:"errorDetail"`
+}
+
+type Options struct {
+	ImageRepo         string `short:"r" long:"image-repo" required:"false" description:"Private Docker image repo URL" value-name:"URL"`
+	ImageTar          string `short:"p" long:"image-tar" required:"false" description:"Path to image tar" value-name:"PATH"`
+	//ImageRepoUser     string `long:"repo-user" required:"false" description:"Private Docker image repo URL" value-name:"USER"`
+	//ImageRepoPw       string `long:"repo-pw" required:"false" description:"Private Docker image repo URL" value-name:"PASSWORD"`
```
jharrod, line 257: These will come back if I can get the go-docker package working with image push. 
It is making a system call to push, currently.

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+const (
+	NatsImageName               = "nats"
+	ConnectorDefaultsConfigPath = "./astraconnector_defaults.yaml"
+	YamlOutputPath              = "./deployConfig.yaml"
+	OperatorYamlPath            = "./astraconnector_operator.yaml"
+	ConnectorNamespace 			= "astra-connector"
+	ConnectorOperatorNamespace  = "astra-connector-operator"
+)
+
+func main() {
```
jharrod, line 278: There could definitely be a lot more functions being used. I've been focused on just getting it working under the time crunch. I would like to come back and make this prettier when I can

astraconnector_defaults.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: netapp.astraconnector.com/v1
```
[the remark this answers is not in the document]
raghu-sarangapani: There is already a copy at the top level that @bmason42 created. astra-connector-operator/agentconfig.yaml

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+		}).Info("Loaded image")
+	}
+	return imageList, nil
+}
+
+func tagImages(dockerClient *client.Client, images []string, repoPrefix string) ([]string, error) {
+	var imagesWithRepo []string
+	for _, image := range images {
+		//todo: update operator to leave the repo in place)E.g. 'myPrivateRepo/asdf.com/theotw/echo-proxylet:1.2.3'
+		newTag := fmt.Sprintf("%s/%s", repoPrefix, strings.TrimPrefix(image, "theotw/"))
```
bmason42, line 141: probably fine for now, but we may load from other places in the future

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+	output, err := createNamespace(ConnectorNamespace)
+	checkFatalErr(err)
+	log.Info(output)
+
+	log.Info("Creating Astra Controller Operator namespace")
+	output, err = createNamespace(ConnectorOperatorNamespace)
+	checkFatalErr(err)
+	log.Info(output)
+
+	// Install Astra Controller Operator
```
raghu-sarangapani, line 366: There are multiple places where it says Astra Controller. Did you mean to say Astra Connector?

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+		checkFatalErr(err)
+
+		log.Info("Tagging images")
+		taggedImages, err := tagImages(dockerClient, loadedImages, opts.ImageRepo)
+		checkFatalErr(err)
+
+		//todo: Fix the auth issue preventing this from succeeding
+		//log.Info("Pushing images")
+		//err = pushImages(dockerClient, &opts, taggedImages)
+		//checkFatalErr(err)
```
raghu-sarangapani, line 327: If the docker push is not working, then the pods wont come up right? Is the push a manual operation currently?

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	// input check
+	if opts.HostAlias {
+		if opts.HostAliasIP == "" {
+			log.Fatal("error: 'hostAliasIP' must be defined if the 'hostAlias' flag is set ")
+		}
+	}
+
+	if opts.ImageTar != "" && opts.ImageRepo != "" {
```
bmason42, line 312: Is this already tested for?
Also, for these sort of tests, best to trim off spaces.  And maybe check len

bmason42 (approved):

installer/install.go
jharrod: Agreed. It will need a tweak in the operator

installer/install.go
jharrod: Ah yeah, you're right. I just added a check for this above. I'll clean that up

I don't think trim is needed in this case. These args are coming from the go-flags package which will trim the whitespace unless the user escapes it or quotes it (how CLIs do)>

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+		dockerClient, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+		checkFatalErr(err)
+
+		log.Info("Loading image tar")
+		loadedImages, err := loadImageTar(dockerClient, opts)
```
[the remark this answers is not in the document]
jharrod: Yes and yes :)

installer/install.go
jharrod: No, push is below. It's doing it via system call instead of using this package i have commented out

installer/install.go
jharrod: Oops. Yes. Thanks

installer/install.go
```diff
@@ -0,0 +1,397 @@
...
+	log.Info(output)
+
+	log.Info("Creating Astra Controller Operator namespace")
+	output, err = createNamespace(ConnectorOperatorNamespace)
+	checkFatalErr(err)
+	log.Info(output)
+
+	// Install Astra Controller Operator
+	operatorYamlPath, err := filepath.Abs(OperatorYamlPath)
+	applyCmd := exec.Command("kubectl", "apply", "-n", ConnectorOperatorNamespace, "-f", operatorYamlPath)
```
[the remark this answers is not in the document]
jharrod: If KUBECONFIG is set, it will use that. Is there something more we need?

installer/install.go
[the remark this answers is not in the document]
bmason42: cool

installer/install.go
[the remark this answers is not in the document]
jharrod: Actually. I think i see what you mean now. I'll pass it along with the cmd. Good catch

PR #10, opened 2022-03-08
Add Security Best Practices section to README

coffeefreak101:
Add Security Best Practices section in README, to inform users of ways they can improve the security of the connector post deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,46 @@ kubectl delete -f config/samples/astraconnector_v1.yaml -n astra-connector
 ```
 kubectl delete -f astraconnector_operator.yaml -n astra-connector-operator
 ```
+
+## Security Best Practices
+### Use RBAC
+Use RBAC (Role Based Access Control) in Kubernetes to limit access to the connector namespace after deployment
+
+### Use a service mesh
+Use a service mesh to secure communications between pods
+
+### Limit token usage for registration
+The API token used for registration is only used once, and is not used after registering the connector.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +40,46 @@ kubectl delete -f config/samples/astraconnector_v1.yaml -n astra-connector
...
+### Use RBAC
+Use RBAC (Role Based Access Control) in Kubernetes to limit access to the connector namespace after deployment
+
+### Use a service mesh
+Use a service mesh to secure communications between pods
+
+### Limit token usage for registration
+The API token used for registration is only used once, and is not used after registering the connector.
+If the token was created specifically for registering the connector, we recommend discarding it after registration is complete.
+The token is not use by the connector after registration.
```
raghu-sarangapani, line 54: I would say remove it from the CRD and revoke it from Astra. If its still in the CRD and the operator reconciles due to some change event, it will keep complaining about an invalid token. From the private cluster side, there is no way to tell if the locationID has been set in the cloud-extensions DB without making an API call.

jharrod (approved):
lgtm

README.md
coffeefreak101: Good catch. I've updated the documentation.

raghu-sarangapani (approved):

bmason42 (approved):
this is all cool.  It would be nice if we can get a test with one or two service meshes (meshI?)
Perhaps we should be setting up the RBAC roles with the operator.  I believe we register some alredy

PR #12, opened 2022-03-09
Add install bundle git action

jharrod:
Additions to github actions:

When building on master:
* Builds an install bundle with install exe and image tar
* Creates a Github Draft Release, using the install bundle as an artifact there
* Install bundle is versioned the same as the operator

In order to create the full release, one of us just needs to navigate to the releases page and click the button.

Example Draft Release: https://github.com/NetApp/astra-connector-operator/releases

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -72,19 +72,19 @@ jobs:
           git push --set-upstream origin master
 
       - name: Build Install Bundle
-#        if: github.ref == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master'
         run: |
           export VERSION=$(cat VERSION)
           make install-bundle
 
       - name: Get Version
-#        if: github.ref == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master'
         id: vars
         run:
           echo ::set-output name=version::$(cat VERSION)
 
       - name: Create Release
-#        if: github.ref == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master'
         id: create_release
         uses: actions/create-release@v1
         env:
@@ -97,7 +97,7 @@ jobs:
           prerelease: false
 
       - name: Upload Release Image Bundle
-#        if: github.ref == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master'
         id: upload-release-asset
         uses: actions/upload-release-asset@v1
         env:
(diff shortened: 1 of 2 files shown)


raghu-sarangapani (approved):

.github/workflows/cicd.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build_and_test:
-    name: CICD tests
```
bmason42, line 13: Did changing this name mess up anyones work flow?

.github/workflows/cicd.yml
```diff
@@ -70,3 +70,43 @@ jobs:
...
+        if: github.ref == 'refs/heads/master'
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG_VERSION: "v${{ steps.vars.outputs.version }}"
+        with:
+          tag_name: ${{ env.TAG_VERSION }}
+          release_name: Release ${{ env.TAG_VERSION }}
+          draft: true
```
bmason42, line 96: You know, on reflection we wont remember to hit the button.  Lets just go for it

Makefile
```diff
@@ -245,4 +242,17 @@ install-exe-linux-amd: install-exe
 install-exe:
 	rm -rf $(OUTPUT_INSTALL_EXE_DIR)
 	mkdir -p $(OUTPUT_INSTALL_EXE_DIR)
-	cd $(INSTALL_DIR) && go build -ldflags "-X github.com/NetApp/astra-connector-operator/installer/install.VERSION=${BUILD_VERSION}" -v -o ${OUTPUT_INSTALL_EXE_DIR}/install-${GOARCH}-${GOOS} ${INSTALL_DIR}/install.go
+	cd $(INSTALL_DIR) && go build -ldflags "-X github.com/NetApp/astra-connector-operator/installer/install.VERSION=${VERSION}" -v -o ${OUTPUT_INSTALL_EXE_DIR}/install-${VERSION}-${GOARCH}-${GOOS} ${INSTALL_DIR}/install.go
```
bmason42, line 245: I think we use underscores instead of dashes in most file names.  Can you confirm?

bmason42, line 245: Does go put a .exe by default or do we have to handle that? (For Windows)

bmason42 (approved):

.github/workflows/cicd.yml
jharrod: Good question. Afaik this just changes what is displayed in the githib UI. @raghu-sarangapani do you know if this is ok to change?

Makefile
jharrod: I think so. Switching to underscore.

.github/workflows/cicd.yml
raghu-sarangapani: It should be fine, we'll find out :)

Makefile
jharrod: It does not. I will add it manually

PR #13, opened 2022-03-16
change the cloud bridge default url to staging

raghu-sarangapani:
Change the cloud bridge default url to staging


bmason42 (approved):
lgtm

PR #17, opened 2023-01-17
updated the version and made the cluster name optional

bmason42:
updated the version and made the cluster name optional


version.txt
```diff
@@ -1 +1 @@
-1.0
\ No newline at end of file
+2.0
```
coffeefreak101, line 1: Do we need a major version change? I would think that making a required field optional would only need a minor version change. I don't really care too much though.

coffeefreak101 (approved):

astraconnector_operator.yaml
```diff
@@ -462,7 +461,7 @@ spec:
         - --leader-elect
         command:
         - /manager
-        image: netapp/astra-connector-operator:1.0.202209131709
+        image: netapp/astra-connector-operator:0.0.1
```
raghu-sarangapani, line 464: is this intentional?

version.txt
bmason42: Do we ?  Probably not.  I am planning on a few more changes and just thought to bump it to 2

astraconnector_operator.yaml
bmason42: no, not sure how that happened.  In fact, it should go to 2.0

bmason42: the regenerate script sets that back to 0.0.1

bmason42: fixed now

PR #21, opened 2023-03-02
Add a status field for operator crd

raghu-sarangapani:
Add a status field for operator crd

### DIFF
--- a/api/v1/astraagent_types.go
+++ b/api/v1/astraagent_types.go
@@ -71,12 +71,14 @@ type AstraConnectorStatus struct {
 type NatssyncClientStatus struct {
 	Registered       string `json:"registered"`
 	AstraConnectorID string `json:"astraConnectorID"`
+	Status           string `json:"status"`
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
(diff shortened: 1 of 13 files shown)


api/v1/astraagent_types.go
```diff
@@ -71,12 +71,14 @@ type AstraConnectorStatus struct {
...
 	Registered       string `json:"registered"`
 	AstraConnectorID string `json:"astraConnectorID"`
+	Status           string `json:"status"`
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name="Registered",type=string,JSONPath=`.status.natssync-client.registered`
 //+kubebuilder:printcolumn:name="AstraConnectorID",type=string,JSONPath=`.status.natssync-client.astraConnectorID`
+//+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.natssync-client.status`
```
vivintw, line 81: nice

bmason42 (approved):
The one thing I would say is it would be nice, at some point, if we could have all the status strings in 1 spot, so we can see what all the status could be, which would make it easier to write test for them

bmason42 (approved):

PR #27, opened 2023-05-26
ASTRACTL-27146 - adds sample CR to bundle yaml

colep-ntap:
Ignore the branch name, I'm going to handle versioning in a separate PR.

This adds a sample operator CR to the release bundle that is created on pushes to master. This is required for downstream CICD pipelines that will want to consume a bundle that contains all of the pieces it needs to deploy the connector operator.

We may also want to add a convenience script for deploying that replaces the fields in the sample CR with params provided by the user. TBD

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,6 +257,7 @@ bundle-base:
 	cp ${OUTPUT_IMAGE_TAR_DIR}/astra-connector-images.tar $(INSTALL_BUNDLE_DIR)
 	cp ${MAKEFILE_DIR}/controllerconfig.yaml $(INSTALL_BUNDLE_DIR)/controllerconfig.yaml
 	cp ${MAKEFILE_DIR}/astraconnector_operator.yaml $(INSTALL_BUNDLE_DIR)/astraconnector_operator.yaml
+	cp ${MAKEFILE_DIR}/config/samples/astraconnector_v1.yaml $(INSTALL_BUNDLE_DIR)/astraconnector_v1.yaml
 
 
 install-bundle: image-tar install-exes bundle-base


Makefile
```diff
@@ -257,6 +257,8 @@ bundle-base:
 	cp ${OUTPUT_IMAGE_TAR_DIR}/astra-connector-images.tar $(INSTALL_BUNDLE_DIR)
 	cp ${MAKEFILE_DIR}/controllerconfig.yaml $(INSTALL_BUNDLE_DIR)/controllerconfig.yaml
 	cp ${MAKEFILE_DIR}/astraconnector_operator.yaml $(INSTALL_BUNDLE_DIR)/astraconnector_operator.yaml
+	cp ${MAKEFILE_DIR}/config/samples/astraconnector_v1.yaml $(INSTALL_BUNDLE_DIR)/astraconnector_v1.yaml
```
bmason42, line 260: I guess I am confused of why we are doing this?  
There is a astraconnector.yaml in the root directory that is updated with the right version of the connector with each build.
What is this going to do?  And does it get the replacement of the image tag?

bmason42 (changes requested):

Makefile
colep-ntap: It allows our downstream CICD pipelines to have one bundle to pull, instead of having to make a series of requests to grab all of the files required to install the connector.
It will also ensure that you are getting the correct version of the operator yaml.

colep-ntap: Oh I'm just noticing the controllerconfig file already being copied in here. Ok I'll delete this line

bmason42 (approved):

jharrod (approved):

PR #37, opened 2023-06-13
update make tar

jharrod:
Updates how we make the tar on the main branch due to files moving location


astraconnector_operator.yaml
```diff
@@ -1,496 +0,0 @@
-apiVersion: v1
```
jharrod, line 1: I added this thinking this was the original issue. 
This file was moved to details/operator-sdk/astraconnector_operator.yaml - so we don't actually need this one here.

colep-ntap (approved):

PR #47, opened 2023-06-15
Remove git pull cicd

jharrod:
Remove git pull in cicd code, causes issue seen here:
https://github.com/NetApp/astra-connector-operator/actions/runs/5281187125/jobs/9554828115

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -77,7 +77,6 @@ jobs:
         run: |
           export VERSION=$(cat VERSION)
           export BRANCH=$(cat BRANCH)
-          git pull --rebase
           make generate-operator-yaml
           git config --global user.name ${{ secrets.ACTIONS_USER }}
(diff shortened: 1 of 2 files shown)


colep-ntap (approved):

.github/workflows/cicd.yml
```diff
@@ -77,7 +77,6 @@ jobs:
         run: |
           export VERSION=$(cat VERSION)
           export BRANCH=$(cat BRANCH)
-          git pull --rebase
           make generate-operator-yaml
           git config --global user.name ${{ secrets.ACTIONS_USER }}
```
bmason42, line 81: Why did we make the user and email secrets?

bmason42 (approved):

.github/workflows/cicd.yml
jharrod: Not sure, that was @raghu-sarangapani, a lot of this code stayed the same

PR #50, opened 2023-06-16
Updating Default of Images

oscarr-ntap:
Updating default images so it matches what is in docker.io and in my docker.repo.eng.netapp.com/oscarr/arch3.0. Now the users won't need to specify images unless they want to test a different non default image. And the only difference between docker.io and internal image is you'll need a secret to pull connector image from docker.io

### DIFF
--- a/common/constants.go
+++ b/common/constants.go
@@ -5,18 +5,18 @@
 package common
 
 const (
-	DefaultImageRegistry = "theotw"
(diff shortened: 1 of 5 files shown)


bmason42 (approved):
Before production we will want to change the default Image registry to cr.astra.netapp.io

PR #52, opened 2023-06-21
Astractl 27157 prechecks

colep-ntap:
Adds a k8s version precheck. Ensures the version of the cluster being deployed on is between 1.24 and 1.27


details/k8s/precheck/precheck.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2023 NetApp, Inc. All Rights Reserved.
```
kchidrala, line 1: We have this file `details/operator-sdk/api/v1/astraconnector_validator.go`, I think we should add all these validations here

colep-ntap: I would argue this falls out of the scope of the validator file. That seems specific to customer inputs. Things like platform checks, version checks, etc. requires more of a framework. 

Keep in mind also, this is going to expand in scope greatly once we begin defining rules for specific platforms.

colep-ntap: One other thought about this: I think for k8s specific validations, they need to occur outside of that validation file so we don't create a circular dependency between the k8s package and the connector package.

kchidrala (approved):

PR #60, opened 2023-06-27
Disable deploying to default namespace

colep-ntap:
This adds a validation that disallows deploying to the default namespace.


kchidrala (approved):

colep-ntap:
> Looks good but it does bring unit test down a whole 1%. haha I'm fine with merging and adding later on though

Oh good call, I'll add a test for this

PR #72, opened 2023-08-03
ASTRACTL-28473 Connector operator to send neptuneV1 capability conditionally

kchidrala:
When Users install connector-operator, with ACOP_FEATUREFLAGS_DEPLOYNEPTUNE=true env var, connector operator should be sending an additional connector capability `neptuneV1` to astra

### DIFF
--- a/common/constants.go
+++ b/common/constants.go
@@ -4,6 +4,8 @@
 
 package common
 
+import "github.com/NetApp-Polaris/astra-connector-operator/app/conf"
+
 const (
 	DefaultImageRegistry = "netappdownloads.jfrog.io/docker-astra-control-staging/arch30/neptune"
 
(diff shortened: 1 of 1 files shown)


common/constants.go
```diff
@@ -51,6 +53,7 @@ const (
 	AstraPrivateCloudType = "private"
 	AstraPrivateCloudName = "private"
 
+	ConnectorNeptuneCapability = "neptuneV1"
```
oscarr-ntap, line 56: Does this need a flag for 23.10 or it should be fine as it is?

oscarr-ntap (approved):

common/constants.go
kchidrala: I am using this flag conditionally only when Deploy Neptune Feature Flag is set to true

PR #149, opened 2023-10-10
modify requeues

colep-ntap:
Modifies requeue timeouts depending on error case. We want to timeout for some configurable time if there is an error to allow the user to view the error condition.


details/operator-sdk/controllers/astraconnector_controller.go
```diff
@@ -151,7 +155,7 @@ func (r *AstraConnectorController) Reconcile(ctx context.Context, req ctrl.Reque
 
 	// POST/PUT Managed Cluster
 	_ = r.updateAstraConnectorStatus(ctx, astraConnector, natsSyncClientStatus)
-	return ctrl.Result{}, nil
+	return ctrl.Result{RequeueAfter: time.Minute * conf.Config.ErrorTimeout()}, nil
```
kchidrala, line 158: This is when the reconciliation ends without any issues, should be no requeue @colep-ntap

PR #199, opened 2023-11-21
ASTRACTL-29245: update for arch-3.0

jharrod:
These are changes to support the "cluster add assist" workflow, which means using the UI to create a "dummy"/pending cluster, and then generating commands for the user to apply to their app cluster to install connector-operator.

The new thing here is that when connector is installing, there will be a clusterID available but Astra will have no ability to talk to the cluster yet, so there is no way to get storage classes.

The changes here make it so, if "connectorInstall" is set in Astra Cluster API, the operator will not attempt to GET or set a default storage class. 
Also, if "connectorInstall" is set, it will update its value to "installed" upon PUT/POST managedClusters


app/register/register.go
```diff
@@ -95,11 +96,11 @@ type ClusterRegisterUtil interface {
...
 	UpdateCluster(astraHost, cloudId, clusterId, astraConnectorId, apiToken string) error
-	CreateOrUpdateCluster(astraHost, cloudId, clusterId, astraConnectorId, clustersMethod, apiToken string) (ClusterInfo, error)
+	CreateOrUpdateCluster(astraHost, cloudId, clusterId, astraConnectorId, connectorInstall, clustersMethod, apiToken string) (ClusterInfo, error)
 	GetStorageClass(astraHost, cloudId, clusterId, apiToken string) (string, error)
-	CreateManagedCluster(astraHost, cloudId, clusterID, storageClass, apiToken string) error
-	UpdateManagedCluster(astraHost, clusterId, astraConnectorId, apiToken string) error
-	CreateOrUpdateManagedCluster(astraHost, cloudId, clusterId, astraConnectorId, managedClustersMethod, apiToken string) (ClusterInfo, error)
+	CreateManagedCluster(astraHost, cloudId, clusterID, storageClass, connectorInstall, apiToken string) error
+	UpdateManagedCluster(astraHost, clusterId, astraConnectorId, connectorInstall, apiToken string) error
+	CreateOrUpdateManagedCluster(astraHost, cloudId, clusterId, astraConnectorId, connectorInstall, managedClustersMethod, apiToken string) (ClusterInfo, error)
```
jharrod, line 103: Most of the changes here are to take "connectorInstall" into consideration. 
If connectorInstall is set - this means we are in the new arch-3.0 flow using the dummy cluster.

This means we want to skip Getting and Setting the default SC when this is set. Otherwise we want to do things the old way

palicena (approved):

app/register/register.go
```diff
@@ -899,16 +910,25 @@ func (c clusterRegisterUtil) CreateOrUpdateManagedCluster(astraHost, cloudId, cl
...
+			// If connectorInstall is not set, this is the old way of installing. We should be able to get the SC from Astra
+			storageClass, err = c.GetStorageClass(astraHost, cloudId, clusterId, apiToken)
+			if err != nil {
+				return ClusterInfo{}, err
+			}
+		} else {
+			// This is the new way of installing for arch-3.0. This way updates a pending cluster already created in Astra.
+			// There is no way to get the SCs from Astra, nor is there a way for Astra to set a default SC in arch 3.0
+			c.Log.Info("Arch-3.0 cluster detected, skipping setting storage class")
+			storageClass = ""
```
jharrod, line 926: This is where we skip getting the SC for arch-3.0 clusters

oscarr-ntap (approved):

PR #203, opened 2023-11-30
Set connectorInstall="pending" for new cluster-add workflow

jharrod:
Include `connectorInstall="pending"` if connector-operator does a POST /clusters. There are some checks in Astra side that key off of this value for the new workflows, specifically this will cause the kubeconfig generation to be skipped for this arch3.0 cluster, and cause parts of the old cluster add/discovery to be skipped.


colep-ntap (approved):

app/register/register.go
```diff
@@ -100,7 +101,7 @@ type ClusterRegisterUtil interface {
 	GetStorageClass(astraHost, cloudId, clusterId, apiToken string) (string, error)
 	CreateManagedCluster(astraHost, cloudId, clusterID, storageClass, connectorInstall, apiToken string) error
 	UpdateManagedCluster(astraHost, clusterId, astraConnectorId, connectorInstall, apiToken string) error
-	CreateOrUpdateManagedCluster(astraHost, cloudId, clusterId, astraConnectorId, connectorInstall, managedClustersMethod, apiToken string) (ClusterInfo, error)
```
jharrod, line 103: Removing connectorInstall from this func. It isn't needed anymore. The unit test changes are due to this

PR #206, opened 2023-12-05
ASTRACTL-30644: replace relayV1 with connectorV2 capability

palicena:
With the k8s-relay being completely removed for this release, this PR is the first step in replacing `relayV1` connector capability with `connectorV2`. This new capability basically means it's an arch3.0 connector. I will be following up with another PR completely removing the RelayV1 capability after merging the [related Polaris PR](https://github.com/NetApp-Polaris/polaris/pull/16625).

BTW, I'm definitely open to a different name for this capability, or other ideas, as it's not the most intuitive in the world (couldn't figure anything better but maybe someone else will).

_Note: I did look into simply removing RelayV1 entirely (i.e. without adding ConnectorV2), but there are a few issues with going down that road; the main one is that we wouldn't be able to reliably detect if this is a arch-3.0-connector-based install or not, which would cause Astra to fall back to thinking it's the old ProxyV1 connector. We can infer it in some cases (such as if the `neptuneV1` capability is present), but in others we unfortunately can't._

Related PR in polaris repo: https://github.com/NetApp-Polaris/polaris/pull/16625


palicena:
Wasn't sure who to have review, so I added the suggested reviewers. Add whoever you think is appropriate if not yourselves!

jharrod (approved):

oscarr-ntap (approved):

PR #216, opened 2023-12-13
Precheck VolumeSnapshotClass

oscarr-ntap:
Precheck VolumeSnapshotClass - Also if we hit pre check errors we can now see them from the status.

There is a high level pre-check flag to skip prechecks if we do not want them. Default is to not skip prechecks

```
apiVersion: astra.netapp.io/v1
kind: AstraConnector
metadata:
  name: astra-connector
  namespace: neptune-system
spec:
  astra:
    accountId: c2446fbb-8ea0-443d-9611-0ebfa3719668
    clusterName: caas2
    skipTLSValidation: true
    tokenRef: astra-token
  natsSyncClient:
    cloudBridgeURL: https://10.193.63.27
  imageRegistry:
    name: netappdownloads.jfrog.io/docker-astra-control-staging/arch30/neptune
    secret: regcred
  skipPreCheck: false
```


kchidrala:
Do we want to expose `skipPreCheck ` flag to customers? If not, we can add it as an environment variable to the operator instead? Thoughts?

details/k8s/k8s_util.go
```diff
@@ -87,3 +91,22 @@ func (r *K8sUtil) VersionGet() (string, error) {
...
+// IsCRDInstalled returns the server version of the k8s cluster.
+func (r *K8sUtil) IsCRDInstalled(crdName string) bool {
+	crd := &apiextv1.CustomResourceDefinition{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: crdName}, crd)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			r.log.V(3).Info(crdName + " CRD does not exist")
+			return false
+		} else {
+			r.log.V(3).Info("Failed to get VolumeSnapshotClass CRD: "+crdName, err)
```
kchidrala, line 104: nit: there should not be a reference to "VolumeSnapshotClass" in the string

kchidrala (approved):

colep-ntap (approved):

PR #224, opened 2023-12-22
ASTRACTL -30644: replace relayV1 with connectorV2 capability (for real this time)

palicena:
With the k8s-relay being completely removed for this release, this PR is the first step in replacing `relayV1` connector capability with `connectorV2`. This new capability basically means it's an arch3.0 connector.

Note: you're not crazy if you're feeling some kind of deja-vu; I merged these changes in a while back but it caused unforeseen issues and got reverted until I could figure things out. Turns out, Polaris PR needs to merge first, and then this one, and then another Polaris PR to completely remove the relayV1 capability.

Related PR in polaris repo: https://github.com/NetApp-Polaris/polaris/pull/16625


common/constants.go
```diff
@@ -57,7 +57,7 @@ const (
 	AstraPrivateCloudName = "private"
 
 	ConnectorNeptuneCapability = "neptuneV1"
-	ConnectorRelayCapability   = "relayV1"
+	ConnectorV2Capability      = "connectorV2" // V2 refers specifically to Arch 3.0 connector and beyond
```
palicena, line 60: I was going to keep the relayV1 capability around previously, but now that we pin the operator version in polaris, I figured it was fine to do it all in one fell swoop instead.

palicena: I lied, going to merge this first, as the Polaris PR will need the new build version for its PR. It might screw over some people for 1-2 hours (those living on the cutting edge as I mentioned), but it's Friday and they should really just be using the pinned version that's in the Polaris repo anyway.

jharrod (approved):

PR #229, opened 2024-01-12
ASTRACTL-30993 adding http status to message

colep-ntap:
Including http status response in our logs and CR status updates instead of just the code. This provides more detail for customers attempting to debug.


app/register/register.go
```diff
@@ -583,7 +585,7 @@ func (c clusterRegisterUtil) GetClusters(astraHost, cloudId, apiToken string) (G
 	}
 
 	if clustersResp.StatusCode != http.StatusOK {
-		return clustersRespJson, fmt.Sprintf("GET call to %v failed with http status code %v", url, clustersResp.StatusCode), errors.New("get clusters failed " + strconv.Itoa(clustersResp.StatusCode))
+		return clustersRespJson, fmt.Sprintf("GET call to %v failed with http status %v", url, clustersResp.Status), errors.New("get clusters failed " + strconv.Itoa(clustersResp.StatusCode))
```
jharrod, line 588: is response.Body not available in all cases? It looks like it is missing from some of these but not sure if that's expected

colep-ntap: I just pushed a change rewriting a bunch of these so they're standardized, so that might clear up your question a bit, but the response body has to be parsed before we can return it. For some reason this check is above the attempt to parse the body though. I'll re-arrange so we have that available

jharrod (approved):
LGTM, thanks Cole

oscarr-ntap (approved):

kchidrala (approved):

PR #231, opened 2024-01-15
remove neptune resources before AstraConnectorCR is deleted

alexkarlis:
We need to try and remove all Neptune resources before continuing on and removing the `AstraConnector` CR. Not only is this good to clean up, it will also help cicd pipelines as well.


details/operator-sdk/controllers/astraconnector_controller.go
```diff
@@ -196,6 +204,38 @@ func (r *AstraConnectorController) Reconcile(ctx context.Context, req ctrl.Reque
...
+	// Get a list of all installed CRDs
+	crdList := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := r.Client.List(ctx, crdList); err != nil {
+		log.Error(err, "Unable to list CRDs")
+		return
+	}
+
+	for _, crd := range crdList.Items {
+		// Only look for the Neptune ones
+		if strings.Contains(crd.Name, "astra") && !strings.Contains(crd.Name, "astraconnectors.astra.netapp.io") {
```
oscarr-ntap, line 220: The contains should be more specific "astra.netapp.io", companies like astrazenca could contain astra crd possibly and this would be deleting those crd.

alexkarlis: Good point

oscarr-ntap (approved):

PR #236, opened 2024-01-19
Remove neptune resources on `AstraConnector` teardown

alexkarlis:
### Description
* create manual list of resources to delete and wait for them to be removed

### DIFF
--- a/details/operator-sdk/controllers/astraconnector_controller.go
+++ b/details/operator-sdk/controllers/astraconnector_controller.go
@@ -21,6 +21,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


details/operator-sdk/controllers/astraconnector_controller.go
```diff
@@ -204,36 +209,136 @@ func (r *AstraConnectorController) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+
+		// Delete all resources of this CRD kind in the namespace AstraConnector is installed
+		if err := r.DynamicClient.Resource(gvr).Namespace(namespace).DeleteCollection(ctx, metav1.DeleteOptions{}, metav1.ListOptions{}); err != nil {
+			log.Error(err, "Unable to delete resources", "Resource", gvr.Resource)
+			return err
+		}
+
+		timeout := time.After(120 * time.Second)
+		ticker := time.NewTicker(5 * time.Second)
```
alexkarlis, line 245: 5 second poll over 120 seconds max

details/operator-sdk/controllers/constants.go
```diff
@@ -0,0 +1,129 @@
+package controllers
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// TODO: need to manually update this list if changes are made
```
oscarr-ntap, line 7: Is there another way of doing this so we don't have to change the cr list in two places. if there was any change in neptune we would then have to change it here too.

oscarr-ntap (approved):

details/operator-sdk/controllers/constants.go
alexkarlis: Yes, technically when we change neptune, we would need to update this list. But adding CRD or bumping versions are few and far between. The purpose for maually defining the list, is so we have control over the ordering of removing CRs as to not to get stuck in a case where the finalizer prevents removal.

PR #238, opened 2024-01-22
Reworked nonRootUser use

litong01:
This PR is to fix the issues deploy astra connector/neptune on openshift cluster.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 # Build the manager binary
-FROM golang:1.21 as builder
+FROM --platform=${BUILDPLATFORM} golang:1.21 as builder
+ARG TARGETOS
+ARG TARGETARCH
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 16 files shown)


Dockerfile
```diff
@@ -1,5 +1,7 @@
 # Build the manager binary
-FROM golang:1.20.5 as builder
+FROM --platform=${BUILDPLATFORM} golang:1.19 as builder
```
litong01, line 2: This actually fixed the issue for multi-platform build which was broken.

oscarr-ntap: --platform=${BUILDPLATFORM} makes sense but we want 1.20 or 1.21 goLang version

litong01: @oscarr-ntap will change to the 1.21.

oscarr-ntap (approved):
Looks good but neptune.yaml is generated so merge https://github.com/NetApp-Polaris/neptune/pull/542/files first.

PR #245, opened 2024-01-25
ASTRACTL-27990 - Adding ACP Check

oscarr-ntap:
ASTRACTL-27990 - Adding ACP Check, making sure acp is installed in prechecks. You can skip this by PreCheckSkip = True


colep-ntap (approved):

app/acp/acp_util.go
```diff
@@ -0,0 +1,41 @@
...
+	for i, tor := range unstructuredList.Items {
+		torcResources[i] = new(torcV1.TridentOrchestrator)
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(tor.Object, torcResources[i])
+		if err != nil {
+			return false, err
+		}
+	}
+
+	if len(torcResources) == 0 {
+		// This is likely the TORC was removed from the cluster
```
saturner, line 31: Just an FYI, another scenario that could get you here is if Trident is installed via tridentctl instead of the operator. Then we won't have a TORC. I don't think you need to make any changes - just making sure you're aware of that possibility.

saturner (approved):
Nice work!

PR #257, opened 2024-02-01
adding check to ensure status is updated

colep-ntap:
This ensures that the Update has completed before another reconcile starts. It's needed b/c Status.Update() takes a while to be registered with k8s


details/operator-sdk/controllers/astraconnector_controller.go
```diff
@@ -468,3 +479,52 @@ func (r *AstraConnectorController) needsReconcile(connector v1.AstraConnector) b
...
+					"Expected status", astraConnector.Status, "Actual status", current.Status)
+				return false, nil
+			}
+
+			// Otherwise stop polling
+			return true, nil
+		}), fmt.Sprintf("AstraConnector status is not updated even after %s", timeout))
+
+	if err == nil {
+		log.Info("Updated AstraConnector status successfully")
```
dseltenr-netapp, line 527: This function doesn't actually update the status so probably should say Status update was reflected in k8s or something? In case this gets used elsewhere that doesn't immediately follow an update call?

dseltenr-netapp (approved):

PR #295, opened 2024-04-22
ASTRACTL-32890:manually update asup tag file

khicks720:
The GitHub action to update this file from neptune is no longer working, so manually updating to match the change in this PR:
https://github.com/NetApp-Polaris/neptune/pull/680

Bug for automatic update here:  [ASTRACTL-32973](https://jira.ngage.netapp.com/browse/ASTRACTL-32973)


khicks720:
Bug already fixed which updated the file, so no need for this PR>

PR #297, opened 2024-04-29
ASTRACTL-32896: tmp kustomization for install script

palicena:
This PR adds a temporary `kustomization.yaml` that the unified installer script will use for some time until we re-organize the deploy YAML and implement neptune helm install as part of the operator.

Corresponding Neptune PR for the `neptune.yaml` folder change: https://github.com/NetApp-Polaris/neptune/pull/696


cluster-install/operator-proxy-role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: operator-proxy-role
+rules:
+  - apiGroups: ["*"]
+    resources: ["*"]
+    verbs: ["*"]
```
palicena, line 8: This is only necessary until some of the RBAC YAML gets re-organized/figured out (which Nathan is working on).

Makefile
```diff
@@ -227,7 +227,7 @@ release: kustomize
 	mkdir build
 	cd details/operator-sdk/config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	cd details/operator-sdk && $(KUSTOMIZE) build config/default > $(BUILD_DIR)/only_astraconnector_operator.yaml
-	cat $(MAKEFILE_DIR)/details/operator-sdk/config/samples/neptune.yaml > $(BUILD_DIR)/astraconnector_operator.yaml
+	cat $(MAKEFILE_DIR)/details/operator-sdk/config/samples-neptune/neptune.yaml > $(BUILD_DIR)/astraconnector_operator.yaml
```
palicena, line 230: This `neptune.yaml` was in a folder that already included a `kustomization.yaml` but wasn't actually included as a resource, giving me two options:
1. Add `neptune.yaml` as a resource to the original kustomization file and potentially disrupt whatever was making use of that kustomization file
2. Move `neptune.yaml` to a new folder so that I can create a new kustomization file

I chose #2 since it didn't seem like anything other than this Make target was making use of `neptune.yaml` right now. LMK if that assumption was wrong.

cluster-install/README.md
```diff
@@ -0,0 +1,4 @@
+This folder is where the Astra cluster install script will live.
+- `kustomization.yaml`: this is a temporary kustomization pulled in by the install 
+script to install the neptune CRDs. It will be removed once the operator is updated to
+install them via Helm.
```
palicena, line 4: Could def use some suggestions/opinions on the folder name, or where else to host the unified installer script (note: this is what customers will be downloading).

Makefile
nhammern: I think moving it out of the /samples directory was the right move here as I am not sure why it was placed there to begin with. One suggestion would be to just call the new directory "neptune" as it doesn't contain any sample CRs. One other place that will need to be updated is this: https://github.com/NetApp-Polaris/neptune/blob/main/.github/workflows/post-merge.yaml#L103

palicena: Neptune PR: https://github.com/NetApp-Polaris/neptune/pull/696

nhammern (approved):
additions look good. just make sure we don't temporarily break the neptune post-merge automation which may expect that neptune.yaml file to be in a specific location

details/operator-sdk/config/manager/manager.yaml
```diff
@@ -1,10 +1,3 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  labels:
-    control-plane: controller-manager
-  name: system
----
```
palicena, line 7: Forgot to mention why this is being removed; it's clashing with the namespace resource in `neptune.yaml`, and probably shouldn't be in here anyhow.

PR #314, opened 2024-05-14
ASTRACTL-33425: opportunistic readme change to force new release

palicena:
A small README change for two reasons:
- The last release didn't contain the unified install script, which would break the cluster add instruction changes I'm planning to make in polaris
    - the release not containing the script might be due to a bug with the github action `actions/create-release`; see https://github.com/actions/create-release/issues/111
- Felt it made sense to align with the folder name (cluster-install) instead of calling the script Unified Installer Script even though there's no mentioned of a "unified installer" anywhere. I like both names, but I think **Cluster Install Script** is probably better from a customer perspective as it's more descriptive of its purpose. That said, would love suggestions or opinions on this.


palicena:
Automation build just came through, solving the main issue I was trying to fix with this PR. Gonna close and we can make README changes closer to release.

PR #315, opened 2024-05-15
ASTRACTL-32772: jq fix, reduce resource limit prompts

palicena:
This PR makes a few small changes to the cluster install script:
1. Fix a `jq` issue that was occurring with lower versions of the tool
2. Properly make the example config file an exception in `gitignore`
3. Reduce the number of prompts for resource limits


.gitignore
```diff
@@ -37,4 +37,4 @@ build/
 
 # Unified Installer: ignore env files that devs may end up keeping around (but whitelist the example)
 cluster-install/*.env
-cluster-install/install-example-config.env
+!cluster-install/install-example-config.env
```
palicena, line 40: The intent was to make this an exception but forgot the `!` when I ported over the gitignore changes.

cluster-install/install.sh
```diff
@@ -1189,7 +1188,7 @@ create_kubectl_patch_for_containers() {
     [ -z "$resource_path" ] && fatal "no resource_path given"
     [ -z "$patch_value" ] && fatal "no patch_value given"
     if ! echo "$resource" | grep -q -- "/"; then fatal "invalid format for given resource '$resource': use 'kind/resource_name'"; fi
-    if ! echo "$patch_value" | jq &> /dev/null; then fatal "given patch_value '$patch_value' is not valid JSON"; fi
+    if ! echo "$patch_value" | jq '.' &> /dev/null; then fatal "given patch_value '$patch_value' is not valid JSON"; fi
```
palicena, line 1191: This is the `jq` change. In lower versions of JQ (observed at version 1.5.1 on Ajay's testing setup; my version is 1.7.1), a naked `jq` command simply prints the help/usage message, while later versions pretty-print (and as a side effect, validate) the piped input.

cluster-install/install.sh
```diff
@@ -1627,41 +1626,18 @@ step_check_kubeconfig_choice() {
 
 # step_query_user_for_resource_count is used to query the user on the number of nodes/namespaces they expect
 # to have on their cluster, which will help us recommend the appropriate resource limit preset.
-step_query_user_for_resource_count() {
```
palicena, line 1630: After talking with @nhammern, we agreed to reduce some of the prompting for resource limits since there was quite a bit.  This will remove the confirmation for node/namespace detection, instead going directly to the recommendation. Might further reduce them later on, but for now this is probably fine.

jharrod (approved):

palicena:
Got Ajay to test the changes and it worked, so i'm gonna go ahead and merge!

PR #320, opened 2024-05-16
ASTRACTL-32772: install script fix for fresh Trident installs when COMPONENTS=TRIDENT_ONLY

palicena:
This PR fixes an issue where Trident did not install as expected when `COMPONENTS=TRIDENT_ONLY` due to the CR apply being hidden behind an `if` condition requiring the Astra Connector to be included in the components.


cluster-install/install.sh
```diff
@@ -583,7 +583,7 @@ prompt_user() {
     if [ "$DISABLE_PROMPTS" == "true" ]; then return 0; fi
 
     while true; do
-        read -p "$prompt_msg" -r "${var_name?}"
+        read -p "${prompt_msg% } " -r "${var_name?}"
```
palicena, line 586: Minor fix to how prompts are displayed; this ensures there will always be a space at the end of the prompt, whether it was manually added by the consumer of the function or not. For example, here's how it would be if the consumer forgot to add the space:
`Please enter the namespace:astra`
With the fix:
`Please enter the namespace: astra`
And if they did include the space, it's a no-op.

cluster-install/install.sh
```diff
@@ -2111,14 +2111,16 @@ step_apply_resources() {
...
         logdebug "apply CRs"
-        if grep -q "AstraConnector" $crs_file_path; then
-            logdebug "delete previous astraconnect if it exists"
-            if ! is_dry_run; then
+        if ! is_dry_run; then
+            if grep -q "AstraConnector" $crs_file_path; then
+                logdebug "delete previous astraconnect if it exists"
                 # Operator doesn't change the astraconnect spec automatically so we need to delete it first (if it exists)
                 kubectl delete -n "$(get_connector_namespace)" deploy/astraconnect &> /dev/null
-                output="$(kubectl apply -f "$crs_file_path")"
```
palicena, line 2119: This line is where we apply CRs, and it was being skipped with `COMPONENTS=TRIDENT_ONLY` since it was hidden behind the 
```
if grep -q "AstraConnector" $crs_file_path; then
```
condition, which will always return false with TRIDENT_ONLY.

cluster-install/install.sh
```diff
@@ -2273,8 +2275,8 @@ step_monitor_deployment_progress() {
 #======================================================================
 #== Main
 #======================================================================
-logln $__INFO "====== Astra Cluster Installer v0.0.1 ======"
 set_log_level
+logln $__INFO "====== Astra Cluster Installer v0.0.1 ======"
```
palicena, line 2279: Moved this line down as it was causing some buggy output due to the log level not being set yet.

jharrod (approved):

josecrosa (approved):

oscarr-ntap (approved):

PR #324, opened 2024-05-19
ASTRACTL-32772: only modify Trident when appropriate

palicena:
This PR will:
- rename `SKIP_TRIDENT_UPGRADE` -> `DO_NOT_MODIFY_EXISTING_TRIDENT`
- the functionality of that env var now extends to ACP, instead of strictly applying to the Trident "upgrade" path
- the previous behavior was confusing testers (and would likely confuse users too) as enabling and updating the ACP image can easily be considered a Trident "upgrade", but in the context of the script, it was separate. This led to existing Trident installs being modified even with SKIP_TRIDENT_UPGRADE=true, subverting user expectation
- the new name hopefully makes things very clear: if you set this to true, the existing Trident install will not be modified in any way, which is probably functionality some users wanted anyhow

Output when it gets blocked:
```
--- Pre-flight check failed! Please resolve the following issues and try again:
* Existing Trident install requires an upgrade but DO_NOT_MODIFY_EXISTING_TRIDENT=true, and no other valid operations can be done due to COMPONENTS=TRIDENT_ONLY.
```

i also tested a fresh Trident install with `COMPONENTS=TRIDENT_ONLY` and `DO_NOT_MODIFY_EXISTING_TRIDENT=true` to make sure it doesn't get blocked with these changes.


cluster-install/install.sh
```diff
@@ -13,6 +13,9 @@ _PROBLEMS=() # Any failed checks will be added to this array and the program wil
 _CONFIG_BUILDER=() # Contains the fully resolved config to be output at the end during a dry run
 _KUBERNETES_VERSION=""
 
+# _TRIDENT_COLLECTION_STEP_CALLED is used as a guardrail to prevent certain functions from being called before
+# existing Trident information (if any) has been collected.
+_TRIDENT_COLLECTION_STEP_CALLED="false"
```
palicena, line 18: This gets set to `true` during `step_collect_existing_trident_info`, and, as the comment mentions, is used to prevent devs from calling certain functions before trident info's been collected as they would've returned bogus values without raising any errors.

cluster-install/install.sh
```diff
@@ -256,13 +259,25 @@ get_connector_namespace() {
...
+    if [ "$_TRIDENT_COLLECTION_STEP_CALLED" != "true" ]; then
+        fatal "this function should not be called until existing Trident information has been collected"
+    fi
     [ -z "$_EXISTING_TRIDENT_NAMESPACE" ] && return 0
     return 1
 }
 
-should_skip_trident_upgrade() {
-    [ "$SKIP_TRIDENT_UPGRADE" == "true" ] && return 0
+trident_will_be_installed_or_modified() {
```
palicena, line 275: With the number of variables involved in deciding whether to do anything with Trident grew, it became harder and harder to understand how it all worked and when to do what. This function is the end-all-be-all "are we doing anything with Trident or not, regardless of what it is" and gates various Trident-related functionality.

cluster-install/install.sh
```diff
@@ -1351,13 +1366,19 @@ step_check_config() {
...
-            add_problem "SKIP_TRIDENT_UPGRADE: required (prompts disabled)" "$longer_msg"
+        if [ -z "$DO_NOT_MODIFY_EXISTING_TRIDENT" ]; then
+            local -r longer_msg="DO_NOT_MODIFY_EXISTING_TRIDENT is required when prompts are disabled."
+            add_problem "DO_NOT_MODIFY_EXISTING_TRIDENT: required (prompts disabled)" "$longer_msg"
         fi
     fi
+    if [ "$DO_NOT_MODIFY_EXISTING_TRIDENT" == "true" ] && [ "$COMPONENTS" == "$__COMPONENTS_ACP_ONLY" ]; then
+        local -r acp_only_err="DO_NOT_MODIFY_EXISTING_TRIDENT: cannot be set to 'true' if using"
+        acp_only_err+=" COMPONENTS='${__COMPONENTS_ACP_ONLY}', as the component in question requires modifying an"
+        acp_only_err+=" existing Trident installation."
```
palicena, line 1377: `DO_NOT_MODIFY_EXISTING_TRIDENT` can't be set to **true** if `COMPONENTS=ACP_ONLY` because that operation, by its very nature, requires modifying an existing Trident install.

`TRIDENT_ONLY` is allowed because it can still mean doing a fresh install. There is a check later on to see if there is an existing Trident install with TRIDENT_ONLY set and DO_NOT_MODIFY_EXISTING_TRIDENT=true, at which point we will error out of the script since that's also incompatible (we just can't know about it until we collect trident info from the cluster).

cluster-install/install.sh
```diff
@@ -256,13 +259,40 @@ get_connector_namespace() {
...
+        fatal "this function should not be called until existing Trident information has been collected"
+    fi
     [ -z "$_EXISTING_TRIDENT_NAMESPACE" ] && return 0
     return 1
 }
 
-should_skip_trident_upgrade() {
-    [ "$SKIP_TRIDENT_UPGRADE" == "true" ] && return 0
+
+trident_will_be_installed_or_modified() {
```
palicena, line 290: With the number of variables involved in deciding whether to do anything with Trident grew, it became harder and harder to understand how it all worked and when to do what. This function is the end-all-be-all "are we doing anything with Trident or not, regardless of what it is" and gates various Trident-related functionality.

cluster-install/install.sh
```diff
@@ -1963,6 +1994,16 @@ step_collect_existing_trident_info() {
     fi
 }
 
+step_existing_trident_flags_compatibility_check() {
```
palicena, line 1997: This is an additional guardrail that was missing even when we had `SKIP_TRIDENT_UPGRADE`. Basically, if the only operations we're supposed to run are Trident-related (due to the chosen COMPONENTS), and there's already an existing Trident install, then this means the only thing we'll do is a Trident upgrade (whether it's the Trident image, operator image, ACP image, or enabling ACP).
If DO_NOT_MODIFY_EXISTING_TRIDENT=true though, then we're being explicitly told we can't do the one thing we're supposed to be doing, so we error out to let the user know of the problem.

The main reason why this is not checked earlier is because we need to collect Trident info before we know whether Trident already exists or not.

cluster-install/install.sh
```diff
@@ -2131,8 +2172,13 @@ step_apply_resources() {
 step_apply_trident_operator_patches() {
     logheader "$__DEBUG" "$(prefix_dryrun "Applying Trident Operator patches...")"
     local -ra patches=("${_PATCHES_TRIDENT_OPERATOR[@]}")
+    local -r patches_len="${#patches[@]}"
 
-    if [ "$LOG_LEVEL" == "$__DEBUG" ] && [ "${#patches[@]}" -gt 0 ]; then
+    if ! trident_will_be_installed_or_modified && [ "$patches_len" -gt 0 ]; then
+        fatal "found $patches_len operator patches (expected 0) despite trident not being installed or modified"
```
palicena, line 2178: Another guardrail to make sure we don't end up generating patches for components that aren't being acted upon.

cluster-install/install.sh
```diff
@@ -2146,8 +2192,13 @@ step_apply_trident_operator_patches() {
 step_apply_torc_patches() {
     logheader "$__DEBUG" "$(prefix_dryrun "Applying TORC patches...")"
     local -ra patches=("${_PATCHES_TORC[@]}")
+    local -r patches_len="${#patches[@]}"
 
-    if [ "$LOG_LEVEL" == "$__DEBUG" ] && [ "${#patches[@]}" -gt 0 ]; then
+    if ! trident_will_be_installed_or_modified && [ "$patches_len" -gt 0 ]; then
+        fatal "found $patches_len torc patches (expected 0) despite trident not being installed or modified"
```
palicena, line 2198: Same guardrail here.

cluster-install/install.sh
```diff
@@ -2327,55 +2378,58 @@ if components_include_connector; then
...
-                loginfo "Trident will not be upgraded."
-            fi
-        # Trident operator upgrade (standalone)
-        elif trident_operator_image_needs_upgraded; then
-            if config_trident_operator_image_is_custom || prompt_user_yes_no "Would you like to upgrade the Trident Operator?"; then
-                step_generate_trident_operator_patch
-            else
-                loginfo "Trident Operator will not be upgraded."
+step_collect_existing_trident_info && exit_if_problems
+step_existing_trident_flags_compatibility_check && exit_if_problems
```
palicena, line 2382: Main changes for this whole Trident-related block is the addition of the "trident flags compatibility" check (i.e. checking if we have a clash between DO_NOT_MODIFY_EXISTING_TRIDENT, COMPONENTS, and an existing Trident install).

The rest is just indentation change from adding the `if trident_will_be_installed_or_modified; then` check on the next line.

cluster-install/install.sh
```diff
@@ -2327,55 +2378,57 @@ if components_include_connector; then
...
+if trident_will_be_installed_or_modified; then
```
palicena, line 2383: Added this check to completely short-circuit any Trident stuff if needed.

cluster-install/install.sh
```diff
@@ -2327,55 +2378,57 @@ if components_include_connector; then
...
+    if trident_is_missing; then
+        step_generate_trident_fresh_install_yaml
+    elif [ -z "$_EXISTING_TRIDENT_OPERATOR_IMAGE" ]; then
+        logwarn "Upgrading Trident without the Trident Operator is not currently supported, skipping."
+    elif existing_trident_can_be_modified; then
```
palicena, line 2388: This is the start of the Trident upgrade/modification section, and it used to be an `else`; now it checks if we're actually allowed to modify the existing Trident install.

jharrod (approved):

PR #325, opened 2024-05-19
ASTRACTL-32772: image check fixes

palicena:
This PR will:
- Swap to a more performant image check (pulling manifest instead of listing tags)
- Improve error output when the curl image check fails with status code `000`
- Add an incremental improvement to deployment monitoring


cluster-install/install.sh
```diff
@@ -36,7 +36,7 @@ _PROCESSED_LABELS=""
 _PROCESSED_RESOURCE_LIMITS=""
 
 # ------------ CONSTANTS ------------
-readonly __RELEASE_VERSION="24.02"
+readonly __RELEASE_VERSION="${RELEASE_VERSION_OVERRIDE:-24.02}"
```
palicena, line 39: Small bonus change for devs: changing this allows you to downgrade Trident (with some little quirks depending on various factors, but I've done it and it mostly works), or upgrade to a non-default version, such as from 23.07 to 23.10 or something like that.

cluster-install/install.sh
```diff
@@ -94,6 +94,11 @@ readonly __WARN=30
 readonly __ERROR=40
 readonly __FATAL=50
 
+# __ERR_FILE should be used when wanting to capture stdout and stderr output of a command separately.
+# You can then use `get_captured_err` to get the captured error. Example:
+#     captured_stdout="$(curl -sS https://bad-url.com 2> "$_ERR_FILE")"
+#     captured_stderr="$(get_captured_err)"
+readonly __ERR_FILE="tmp_last_captured_error.txt"
```
palicena, line 101: From what I could tell, there's no other way to separately capture both stdout/stderr without sending at least one into a file. 

We can generally get away with ignoring **stderr** and instead just check the return code or check if the returned output is empty, but in some cases, the **stderr** output is essential to figuring out what the hell is going on (most usages of curl is one of those cases).  This `__ERR_FILE` is for those situations, and the reason why it's a variable is mainly for consistency across the script.

cluster-install/install.sh
```diff
@@ -493,6 +498,21 @@ get_limits_list_fancy() {
 #----------------------------------------------------------------------
 #-- Util functions
 #----------------------------------------------------------------------
+get_captured_err() {
+    if [ -f "$__ERR_FILE" ]; then
+        cat "$__ERR_FILE"
+        rm -f "$__ERR_FILE" &> /dev/null
```
palicena, line 504: Trying to make this ERR_FILE handling as behind-the-scenes as possible, so it's cleaned up every time we read its value. In the event the script exits before its value is read, there is a cleanup function later on that gets called anyway.

One exception is if we run into a `fatal` error (which ideally will never happen to a customer as those should be reserved for dev-specific guardrails), in which case we deliberately leave the file alone for debugging purposes.

cluster-install/install.sh
```diff
@@ -493,6 +498,21 @@ get_limits_list_fancy() {
...
+    else
+        echo ""
+    fi
+}
+
+debug_is_on() {
```
palicena, line 510: For readability/usability/consistency, since there were a few places where we did this check.

cluster-install/install.sh
```diff
@@ -1005,27 +1025,31 @@ check_if_image_can_be_pulled_via_curl() {
     if [ -z "$image_repo" ]; then fatal "no image_repo given"; fi
     if [ -z "$image_tag" ]; then fatal "no image_tag given"; fi
 
-    local -a args=('-s' '-w' "\n%{http_code}")
+    local -a args=('-sS' '-w' "\n%{http_code}")
```
palicena, line 1028: The `-s` flag makes curl silent, while `-S` makes it output errors to stderr. Combined, it suppresses the normal progress output, but lets errors through.

cluster-install/install.sh
```diff
@@ -1005,27 +1025,31 @@ check_if_image_can_be_pulled_via_curl() {
...
     if [ -n "$encoded_creds" ]; then
         args+=("-H" "Authorization: Basic $encoded_creds")
     fi
+    args+=("-H" "Accept: application/vnd.docker.distribution.manifest.v2+json")
 
-    local -r result="$(curl -X GET "${args[@]}" "https://$registry/v2/$image_repo/tags/list")"
+    local -r result="$(curl -X GET "${args[@]}" "https://$registry/v2/$image_repo/manifests/$image_tag" 2> "$__ERR_FILE")"
```
palicena, line 1034: This is the new way of checking if images exist; we directly fetch the manifest, which is much faster than the previous method of listing tags (especially when there's a lot of tags) when the manifest exists -- something like going from a full minute difference for some images on `docker.repo.eng.netapp.com` to practically instant.

Unfortunately, if the manifest is missing, it takes almost as long as it did with the previous method, for some reason. Still much better though since, in most cases, the manifests _will_ exist.

cluster-install/install.sh
```diff
@@ -1005,27 +1025,31 @@ check_if_image_can_be_pulled_via_curl() {
...
-        #   "name" : "my/repo/astra-connector",
-        #   "tags" : [ "032172b", "074dk52", "1fbc135", "24.02" ]
-        # }
-        if echo "$_return_body" | grep "tags" | grep -q -- "$image_tag"; then
-            return 0
+        return 0
+    elif [ "$_return_status" == 404 ]; then
+        _return_error="the image was not found"
+    elif [ "$_return_status" != 000 ]; then
+        _return_error="$(echo "$_return_body" | jq -r '.errors.[0].message' 2> /dev/null)"
```
palicena, line 1046: This pulls the error message from the body, following the standard JSON path that docker registries return in the body when an error occurs. Can't fully rely on it though, so if we end up empty handed for whatever reason, we just set it to `unknown error` (next few lines).

cluster-install/install.sh
```diff
@@ -1115,13 +1139,24 @@ apply_kubectl_patches() {
...
 
+    local -r sleep_time="5" # Seconds
+    local -r max_checks="$(( (timeout * 60) / sleep_time ))"
+    local counter=0
     loginfo "Waiting on deployment/$deployment (timeout: $timeout)..."
-    if kubectl rollout status -n "$namespace" "deploy/$deployment" --timeout="$timeout" &> /dev/null; then
-        return 0
-    fi
+    while ((counter < max_checks)); do
+        if kubectl rollout status -n "$namespace" "deploy/$deployment" -w=false &> /dev/null; then
```
palicena, line 1150: After testing with just using kubectl rollout's `--timeout` flag, I realized the command will exit early when the deployment in certain failure states, which made the deployment monitoring exit early in many cases since we generally break a few eggs in the process of installing/upgrading things. 

Not sure if there's another option to kubectl rollout that makes it more tolerant, but for now I reverted to using a `while` loop.

cluster-install/install.sh
```diff
@@ -1580,9 +1618,10 @@ step_check_all_images_can_be_pulled() {
             local status="$_return_status"
             local body="$_return_body"
             local error="$_return_error"
+            local problem="[$full_image] image check failed"
```
palicena, line 1621: Just prettying up the output a bit.

cluster-install/install.sh
```diff
@@ -2315,14 +2355,19 @@ step_monitor_deployment_progress() {
     if trident_will_be_installed_or_modified; then
         if is_dry_run; then
             logdebug "skip monitoring trident components because it's a dry run"
-        elif ! wait_for_deployment_running "trident-operator" "$trident_ns" "1m"; then
+        elif ! wait_for_deployment_running "trident-operator" "$trident_ns" "3"; then
             add_problem "trident operator: failed" "The Trident Operator failed to deploy"
-        elif ! wait_for_deployment_running "trident-controller" "$trident_ns" "10m"; then
-            add_problem "trident controller: failed" "Trident failed to deploy"
+        elif ! wait_for_cr_state "torc/$_EXISTING_TORC_NAME" ".status.status" "Installed" "$trident_ns" "12"; then
```
palicena, line 2360: Small change to tracking Trident deployment progress; using **torc** status now instead of the **trident-controller** deployment, the former of which I assume is more reliable since the operator sets it.

cluster-install/install.sh
```diff
@@ -1005,27 +1049,31 @@ check_if_image_can_be_pulled_via_curl() {
     if [ -z "$image_repo" ]; then fatal "no image_repo given"; fi
     if [ -z "$image_tag" ]; then fatal "no image_tag given"; fi
 
-    local -a args=('-s' '-w' "\n%{http_code}")
+    local -a args=('-sS' '-w' "\n%{http_code}")
```
palicena, line 1052: The `-s` flag makes curl silent, while `-S` makes it output errors to stderr. Combined, it suppresses the normal progress output, but lets errors through.

cluster-install/install.sh
```diff
@@ -842,7 +885,8 @@ make_astra_control_request() {
     fi
 
     logdebug "$method --> '$url'"
-    local -r result="$(curl -X "$method" -s $skip_tls_validation_opt -w "\n%{http_code}" "$url" "${headers[@]}")"
+    local -r result="$(curl -X "$method" -sS $skip_tls_validation_opt -w "\n%{http_code}" "$url" "${headers[@]}" 2> "$__ERR_FILE")"
```
palicena, line 888: This is the new way of checking if images exist; we directly fetch the manifest, which is much faster than the previous method of listing tags (especially when there's a lot of tags) when the manifest exists -- something like going from a full minute difference for some images on `docker.repo.eng.netapp.com` to practically instant.

Unfortunately, if the manifest is missing, it takes almost as long as it did with the previous method, for some reason. Still much better though since, in most cases, the manifests _will_ exist.

This is also the first application of the `__ERR_FILE` "pattern". The output of **stdout** gets captured into the `result` variable, while `stderr` is redirected into the __ERR_FILE. We then read the contents of __ERR_FILE on the next line by calling `get_captured_err`, which will **cat** the value and then delete the file.

cluster-install/install.sh
```diff
@@ -1005,27 +1049,31 @@ check_if_image_can_be_pulled_via_curl() {
...
 
-    local -a args=('-s' '-w' "\n%{http_code}")
+    local -a args=('-sS' '-w' "\n%{http_code}")
     if [ -n "$encoded_creds" ]; then
         args+=("-H" "Authorization: Basic $encoded_creds")
     fi
+    args+=("-H" "Accept: application/vnd.docker.distribution.manifest.v2+json")
 
-    local -r result="$(curl -X GET "${args[@]}" "https://$registry/v2/$image_repo/tags/list")"
+    local -r result="$(curl -X GET "${args[@]}" "https://$registry/v2/$image_repo/manifests/$image_tag" 2> "$__ERR_FILE")"
```
palicena, line 1058: This is the new way of checking if images exist; we directly fetch the manifest, which is much faster than the previous method of listing tags (especially when there's a lot of tags) when the manifest exists -- something like going from a full minute difference for some images on `docker.repo.eng.netapp.com` to practically instant.

Unfortunately, if the manifest is missing, it takes almost as long as it did with the previous method, for some reason. Still much better though since, in most cases, the manifests _will_ exist.

cluster-install/install.sh
```diff
@@ -1621,28 +1678,43 @@ step_check_all_images_can_be_pulled() {
...
+    local -r status="$_return_status"
+    local -r body="$_return_body"
+    local err="$_return_error"
+    if [ "$status" == 200 ]; then
         logdebug "astra control: OK"
     else
-        add_problem "astra control: failed ($_return_status)" "Failed to contact Astra Control (status code: $_return_status)"
+        if [ "$status" != 000 ] || [ -z "$err" ]; then err="$(status_code_msg "$status")"; fi
+        logdebug "body: '$body'"
+        add_problem "Failed to contact Astra Control: $err ($status)"
```
palicena, line 1689: This and the few other changes below make use of the new `__ERR_FILE` stderr thing to improve error handling/logging for the Astra Control requests.

cluster-install/install.sh
```diff
@@ -2250,28 +2371,33 @@ step_monitor_deployment_progress() {
     if components_include_connector; then
         if is_dry_run; then
             logdebug "skip monitoring connector components because it's a dry run"
-        elif ! wait_for_deployment_running "operator-controller-manager" "$connector_operator_ns" "1m"; then
+        elif ! wait_for_deployment_running "operator-controller-manager" "$connector_operator_ns" "3"; then
```
palicena, line 2374: With the change to the `wait_for_deployment_running`, we were no longer passing over the timeout value to kubectl for us to parse, so `1m` -> `1` and we just assume the unit is minutes every time. Don't think we really need the granularity of seconds or the scale of hours anyhow.

cluster-install/install.sh
palicena: Hmm, you're right, that's basically all it is until (and if) we change the tagging scheme for the connector/neptune stuff. Will make the change.

jharrod (approved):

PR #326, opened 2024-05-19
ASTRACTL-32772: improvements to TLS validation skips

palicena:
This PR will:
- add a general `SKIP_TLS_VALIDATION` flag, in addition to `CONNECTOR_SKIP_TLS_VALIDATION`. This new flag is now what dictates whether requests in the script will skip TLS validation or not
-  `CONNECTOR_SKIP_TLS_VALIDATION` now defaults to the value of `SKIP_TLS_VALIDATION`, but can be overridden if so desired. It now strictly dictates the value of the connector CR's `skipTLSValidation` flag. 
- If the script runs into an SSL certificate issue, it will now output a hint mentioning `SKIP_TLS_VALIDATION`


cluster-install/install.sh
```diff
@@ -850,6 +866,18 @@ status_code_msg() {
     echo "$status"
 }
 
+add_tls_validation_hint_to_err_if_needed() {
```
palicena, line 869: This is what this ends up looking like:
```
Failed to contact Astra Control at url 'https://10.193.60.83': curl: (60) SSL certificate problem -- try setting SKIP_TLS_VALIDATION=true (WARNING: not for production use!) (000)
```

cluster-install/install.sh
```diff
@@ -1328,7 +1359,7 @@ create_kubectl_patch_for_containers() {
 exit_if_problems() {
     if [ ${#_PROBLEMS[@]} -ne 0 ]; then
         debug_is_on && print_built_config
-        logheader $__ERROR "Pre-flight check failed! Please resolve the following issues and try again:"
+        logheader $__ERROR "Cluster management failed! We've identified the following issues:"
```
palicena, line 1362: Reworded this since it's no longer just a **Pre-flight** check script, and the wording on getting the user to resolve the issues kind of sounded like we're blaming them 😅

jharrod (approved):

PR #328, opened 2024-05-20
ASTRACTL-33565: warn when api token is ingested via config

zachatnetapp:
Throw a warning when a user of the installer script is specifying their API Token in a config file, as that is not secure.


cluster-install/install.sh
```diff
@@ -199,6 +200,13 @@ load_config_from_file_if_given() {
 
     # shellcheck disable=SC1090
     source "$config_file"
+
+    # check if api token was populated after sourcing config file
+    if ["$api_token" != "$ASTRA_API_TOKEN"]; then
+        logwarn "We detected that your ASTRA_API_TOKEN was provided through the CONFIG_FILE, "\
+        "which is insecure! Consider storing it in a safe location and providing it through the command line only when needed."
+    fi
```
palicena, line 208: Just thought of yet another rewording that might be better:
> We detected that your ASTRA_API_TOKEN was provided through the CONFIG_FILE, which may pose a security risk! Make sure to store the configuration file in a safe location, or consider moving the API token out of the file and providing the token through the command line when needed.

palicena (approved):
Tested, worked without issues after the fixes. LGTM!

PR #329, opened 2024-05-21
ASTRACTL-32772: backup TORC, add created-by label, other fixes

palicena:
This PR has a few additions:
- Add `git` as a required tool -- this is indirectly needed due to kustomize requiring it to pull from all CRDs from the astra-connector-operator repo
- Add label `apps.kubernetes.io/created-by: astra-cluster-install-script` to all resources we create
- Add backup of existing TORC if we're upgrading Trident

It also has a few small fixes -- see comments for details.


cluster-install/install.sh
```diff
@@ -39,7 +39,7 @@ _PROCESSED_RESOURCE_LIMITS=""
 readonly __RELEASE_VERSION="24.02"
 readonly __TRIDENT_VERSION="${__TRIDENT_VERSION_OVERRIDE:-"$__RELEASE_VERSION"}"
 
-readonly -a __REQUIRED_TOOLS=("jq" "kubectl" "curl" "grep" "sort" "uniq" "find" "base64" "wc" "awk")
+readonly -a __REQUIRED_TOOLS=("git" "jq" "kubectl" "curl" "grep" "sort" "uniq" "find" "base64" "wc" "awk")
```
palicena, line 42: `git` is required by kustomize to pull all the operator CRDs from the astra-connector-operator repo

cluster-install/install.sh
```diff
@@ -73,6 +73,7 @@ readonly __DEFAULT_TRIDENT_ACP_IMAGE_NAME="trident-acp"
 
 readonly __DEFAULT_CONNECTOR_NAMESPACE="astra-connector"
 readonly __DEFAULT_TRIDENT_NAMESPACE="trident"
+readonly __DEFAULT_TORC_NAME="trident"
```
palicena, line 76: The TORC name is used in a few places, this should make things a bit more consistent for the long term.

cluster-install/install.sh
```diff
@@ -216,13 +217,13 @@ load_config_from_file_if_given() {
 
     # shellcheck disable=SC1090
     source "$config_file"
+    set_log_level
```
palicena, line 220: This fixes an issue where the `logwarn` right below would fail (and output a bunch of errors) because the log level wasn't set yet.

cluster-install/install.sh
```diff
@@ -572,7 +573,7 @@ loginfo() {
 }
 
 logwarn() {
-    log_at_level $__WARN "$1"
+    log_at_level $__WARN "WARNING: $1"
```
palicena, line 576: Felt like the warnings kind of flew under the radar a bit too much. Still not sure about adding this though, so def interested to hear what others think about adding this.

cluster-install/install.sh
```diff
@@ -819,12 +820,11 @@ tool_is_installed() {
...
 
-    if
-        command -v "$tool" &>/dev/null
+    if command -v "$tool" &>/dev/null; then
         return 0
-    then
-        return 1
     fi
+
+    return 1
```
palicena, line 827: I must've been typing with my monitor off when I put this in. The `then` was in the complete wrong place and there was an extra newline for whatever reason. Overall, it made the tool install check mostly useless, which this change should fix (credit to Keir for noticing and testing the fix).

cluster-install/install.sh
```diff
@@ -1369,6 +1369,33 @@ create_kubectl_patch_for_containers() {
...
+    fi
+    backup_name+="${kind}_${resource_name}.yaml"
+
+    local -r resource_name_yaml="$(kubectl get "$kind" "$resource_name" "$namespace_arg" -o yaml 2> $__ERR_FILE)"
+    [ -z "$resource_name_yaml" ] && return 1
+    _return_error="$(get_captured_err)"
+
+    echo "$resource_name_yaml" > "${directory}/${backup_name}"
+    echo "$backup_name"
+}
```
palicena, line 1397: This backup function is used to backup the TORC before/if we apply patches to it.

cluster-install/install.sh
```diff
@@ -1500,8 +1527,14 @@ step_check_config() {
     add_to_config_builder "CONNECTOR_AUTOSUPPORT_ENROLLED"
     add_to_config_builder "CONNECTOR_AUTOSUPPORT_URL"
 
+    # Add our default labels
+    local -r label_indent="    "
+    local -a default_labels=("app.kubernetes.io/created-by=astra-cluster-install-script")
+    _PROCESSED_LABELS="$(process_labels_to_yaml "${default_labels[*]}" "$label_indent")"
```
palicena, line 1533: This adds a label which will allow users (and our pipeline) to identify/select all the resources the script created. It's added alongside any custom user labels:
```
    labels:
      app.kubernetes.io/created-by: astra-cluster-install-script
      testLabel: hello
      testLabelTwo: helloTwo
```

cluster-install/install.sh
```diff
@@ -1716,7 +1749,7 @@ step_check_all_images_can_be_pulled() {
                 #
                 # And so, we'll try to check the image tag using the pull secret credentials (if provided),
                 # and it's great if it succeeds, but it's generally expected to fail.
-                logwarn "* Cannot guarantee the existence of custom Docker Hub image '$full_image', skipping."
+                logwarn "Cannot guarantee the existence of custom Docker Hub image '$full_image', skipping."
```
palicena, line 1752: Cosmetic logging changes here and the next few changes.

cluster-install/install.sh
```diff
@@ -2156,15 +2187,15 @@ step_generate_trident_fresh_install_yaml() {
     local enable_acp="true"
     local labels_field_and_content=""
     if [ -n "$IMAGE_PULL_SECRET" ]; then pull_secret='["'$IMAGE_PULL_SECRET'"]'; fi
-    if [ -n "$PROCESSED_LABELS" ]; then
+    if [ -n "$_PROCESSED_LABELS" ]; then
```
palicena, line 2190: This was missing the underscore `_`, which meant the variable was always empty, which in turn prevented labels from being applied to the TORC.

cluster-install/install.sh
```diff
@@ -2336,6 +2367,18 @@ step_apply_torc_patches() {
...
+    if [ -n "$_EXISTING_TORC_NAME" ]; then
+        local -r backup="$(backup_kubernetes_resource "tridentorchestrator" "$_EXISTING_TORC_NAME" "$__GENERATED_CRS_DIR")"
+        if [ -n "$backup" ]; then
+            loginfo "* Created backup for TridentOrchestrator '$_EXISTING_TORC_NAME': '$backup'"
+        elif [ -n "$_return_error" ]; then
+            logdebug "failed to create backup for TridentOrchestrator '$_EXISTING_TORC_NAME': $_return_error"
+        else
+            logdebug "failed to create backup for TridentOrchestrator '$_EXISTING_TORC_NAME': unknown error"
+        fi
+    fi
```
palicena, line 2380: This is where we take the TORC backup.

cluster-install/install.sh
```diff
@@ -2443,12 +2486,13 @@ step_monitor_deployment_progress() {
         fi
     fi
 
+    local -r torc_name="${_EXISTING_TORC_NAME:-"$__DEFAULT_TORC_NAME"}"
```
palicena, line 2489: Was noticing that the monitoring function was failing the TORC status check sometimes. Turns out that `_EXISTING_TORC_NAME` is empty when, well, there's no existing TORC (duh). So we now fall back to the default TORC name (used for fresh installs) in those cases.

cluster-install/install.sh
```diff
@@ -1095,7 +1095,13 @@ check_if_image_can_be_pulled_via_curl() {
...
         args+=("-k")
     fi
-    args+=("-H" "Accept: application/vnd.docker.distribution.manifest.v2+json")
+    # We accept all formats via '*/*' because we only really care about the status code, but certain multi-platform
+    # images require a more specific format and will return 404 if we only use '*/*', so we add those formats as well
+    local accept_formats="*/*"
+    accept_formats+=", application/vnd.docker.distribution.manifest.list.v1+json"
+    accept_formats+=", application/vnd.docker.distribution.manifest.list.v2+json"
+    accept_formats+=", application/vnd.oci.image.index.v1+json" # Required for ACP
+    args+=("-H" "Accept: $accept_formats")
```
palicena, line 1104: After my recent change to how we do the image checks, @jharrod noticed the `trident-acp` image check was now returning a `404` for whatever reason. Turns out it's because we were missing the format that image manifest uses: `application/vnd.oci.image.index.v1+json`. Added `*/*` and `application/vnd.docker.distribution.manifest.list.v1+json` to boot, just in case.

jharrod (approved):

PR #330, opened 2024-05-21
Astractl 33552 allow default connector and neptune

jharrod:
This PR updates the install script to make the CONNECTOR_IMAGE_TAG and NEPTUNE_IMAGE_TAG params optional. 

I expect most customers to not set these and to allow connector-operator to choose the neptune and connector versions based on its defaults. These defaults are the ones we use in all our our testing and pipelines. Each connector-operator version has a paired connector and neptune version that is fully validated. Any other combination is not vetted/supported. We should probably hide these params from customers and rely on the connector-operator default to enforce known and vetted image combinations.


cluster-install/install.sh
```diff
@@ -365,6 +362,11 @@ config_image_is_custom() {
     local -r component_name="$1"
     local -r default_registry="$2"
     local -r default_base_repo="$3"
+    local -r default_tag="$4"
```
jharrod, line 365: Modified this function so it does not assume all image tags are the same

palicena (approved):
👌

cluster-install/install.sh
```diff
@@ -1129,6 +1132,20 @@ wait_for_deployment_running() {
     [ -z "$deployment" ] && fatal "no deployment name given"
 
     loginfo "Waiting on deployment/$deployment (timeout: $timeout)..."
+
+    # Wait for the deployment to be created. The subsequent 'rollout status' command will fail if called before the
+    # deployment has been created.
+    max_retries=15
```
jharrod, line 1138: I was seeing my connector deployment show up after this check had already failed. Not sure why the deployment was taking so long to appear but adding this check to first wait for the deployment before calling the 'rollout status' command below fixed it

cluster-install/install.sh
```diff
@@ -1637,13 +1642,30 @@ step_check_all_images_can_be_pulled() {
...
+        if config_neptune_image_is_custom; then
+          # Only check neptune image if user has overridden the connector-operator's default.
+          # We do not know the default version and cannot check it due to it being hard coded within the connector-operator image.
+          images_to_check+=("$NEPTUNE_IMAGE_REGISTRY" "$NEPTUNE_IMAGE_REPO" "$NEPTUNE_IMAGE_TAG" "$custom")
+        else
+          # Get the default connector tag
+          file_content=$(curl -sS "https://raw.githubusercontent.com/NetApp/astra-connector-operator/$CONNECTOR_OPERATOR_IMAGE_TAG/common/neptune_manager_tag.txt")
+          # Trim new lines and white space
+          tag="${file_content//[[:space:]]/}"
+          images_to_check+=("$NEPTUNE_IMAGE_REGISTRY" "$NEPTUNE_IMAGE_REPO" "$tag" "$default")
```
palicena, line 1667: (same here)

palicena (approved):

cluster-install/install.sh
```diff
@@ -1637,13 +1642,30 @@ step_check_all_images_can_be_pulled() {
...
-        else images_to_check+=("$default"); fi
+        if config_connector_image_is_custom; then
+          # Only check connector image if user has overridden the connector-operator's default.
+          # We do not know the default version and cannot check it due to it being hard coded within the connector-operator image.
+          images_to_check+=("$CONNECTOR_IMAGE_REGISTRY" "$CONNECTOR_IMAGE_REPO" "$CONNECTOR_IMAGE_TAG" "$custom")
+        else
+          # Get the default connector tag
+          file_content=$(curl -sS "https://raw.githubusercontent.com/NetApp/astra-connector-operator/$CONNECTOR_OPERATOR_IMAGE_TAG/common/connector_version.txt")
+          # Trim new lines and white space
+          tag="${file_content//[[:space:]]/}"
```
palicena, line 1654: Also just realized -- should probably make these two variables (file_content, tag) local to the function (unless I missed their declaration somewhere), as they're gonna end up being global for the whole script. 

Important notes about local bash variables though is that they're scoped to the entire function, even if declared within an `if` block; so if you declare it readonly `local -r my_readonly_var` in multiple places in the same function, you'll probably get an error.

PR #332, opened 2024-05-22
ASTRACTL-33571: check Trident version before trying to enable ACP

zachatnetapp:
check users existing Trident install version to make sure we don't try to enable ACP when it's not supported.


cluster-install/install.sh
```diff
@@ -2087,6 +2088,15 @@ step_collect_existing_trident_info() {
         logdebug "trident image: not found"
     fi
 
+    # Trident version
+    local -r trident_version="$(kubectl get tridentversions -n trident -o json | jq -r '.items.[0].trident_version' 2> /dev/null)"
+    if [ -z "$trident_version" ]; then
+        logwarn "Failed to resolve existing Trident version. ACP may not be supported without an upgrade!"
+    else
+        _EXISTING_TRIDENT_VERSION="$trident_version"
+        logdebug "trident version found: $trident_version"
```
palicena, line 2097: I think it's probably worth making this a `loginfo` TBH, reinforces to the user what their env setup currently is and all. 
```suggestion
        loginfo "* Trident version: $trident_version"
```

cluster-install/install.sh
```diff
@@ -2532,11 +2542,34 @@ if trident_will_be_installed_or_modified; then
...
             # Trident upgrade (includes operator upgrade if needed)
             if trident_image_needs_upgraded; then
-                if config_trident_image_is_custom || prompt_user_yes_no "Would you like to upgrade Trident?"; then
-                    step_generate_torc_patch "$_EXISTING_TORC_NAME" "$(get_config_trident_image)" "" "" "$(get_config_trident_autosupport_image)"
-                    trident_operator_image_needs_upgraded && step_generate_trident_operator_patch
-                else
-                    loginfo "Trident will not be upgraded."
+                # if trident version < 23.10
+                if version_higher_or_equal "23.07" "$_EXISTING_TRIDENT_VERSION"; then
+                    logwarn "Your Trident installation is at version 23.07, while the lowest required version to enable ACP is 23.10."
```
palicena, line 2547: Could the trident version here be lower than 23.07? Might need to put in the actual version in the log:
```suggestion
                    logwarn "Your Trident installation is at version $_EXISTING_TRIDENT_VERSION, while the lowest required version to enable ACP is 23.10."
```

cluster-install/install.sh
```diff
@@ -2532,11 +2542,34 @@ if trident_will_be_installed_or_modified; then
...
+                if ! components_include_connector && (components_include_trident || components_include_acp); then
+                    if prompt_user_yes_no "Would you like to upgrade Trident? Choosing no will exit the script (yes/no):"; then
+                        step_generate_torc_patch "$_EXISTING_TORC_NAME" "$(get_config_trident_image)" "" "" "$(get_config_trident_autosupport_image)"
+                        trident_operator_image_needs_upgraded && step_generate_trident_operator_patch
+                    else
+                        exit 0
+                    fi
+                fi
+
+                if components_include_connector; then
```
palicena, line 2566: Should probably check to see if ACP is already enabled here. It's possible the user has a Trident upgrade available yet already has ACP enabled.

zachatnetapp: Added this below

jharrod (approved):
lgtm

PR #334, opened 2024-05-23
ASTRACTL-32772: prioritize CLI args over cfg file, better errors

palicena:
This PR adds a few things:
- Command line arguments will now take priority over values given through the config file
- Better handling of errors when calling `kubectl get`
- Prevent `trident-operator` resource limits from being applied if the user declined the trident/trident-operator upgrade

I tested these changes through 3 scenarios:
1. Greenfield (no initial connector/trident install)
2. Brownfield (no initial connector, but pre-existing trident at 23.07)
    - For this one, I specifically ran twice, once saying yes to the trident upgrade and once saying no, to make sure the resource limits get applied (or not) appropriately
4. Brownfield, but DO_NOT_MODIFY_EXISTING_TRIDENT=true (same setup as 2nd test)


cluster-install/install.sh
```diff
@@ -39,7 +40,7 @@ _PROCESSED_RESOURCE_LIMITS=""
 readonly __RELEASE_VERSION="24.02"
 readonly __TRIDENT_VERSION="${__TRIDENT_VERSION_OVERRIDE:-"$__RELEASE_VERSION"}"
 
-readonly -a __REQUIRED_TOOLS=("git" "jq" "kubectl" "curl" "grep" "sort" "uniq" "find" "base64" "wc" "awk")
+readonly -a __REQUIRED_TOOLS=("git" "jq" "kubectl" "curl" "grep" "sort" "uniq" "find" "base64" "wc" "awk" "tail" "head")
```
palicena, line 43: `tail` and `head` have been used since the very beginning but got missed when making this list.

cluster-install/install.sh
```diff
@@ -22,6 +22,7 @@ _EXISTING_TRIDENT_IMAGE=""
 _EXISTING_TRIDENT_ACP_ENABLED=""
 _EXISTING_TRIDENT_ACP_IMAGE=""
 _EXISTING_TRIDENT_OPERATOR_IMAGE=""
+_USER_ACCEPTED_TRIDENT_OPERATOR_UPGRADE="" # "true" if yes, "false" if no, "" if the question was never asked
```
palicena, line 25: Currently, resource limits will be applied to the **trident-operator** when an upgrade is possible, even if the user declined said upgrade. This is used to prevent resource limits from being applied to the **trident-operator** if the user declined the upgrade.

cluster-install/install.sh
```diff
@@ -104,6 +105,11 @@ readonly __FATAL=50
 #     captured_stdout="$(curl -sS https://bad-url.com 2> "$_ERR_FILE")"
 #     captured_stderr="$(get_captured_err)"
 readonly __ERR_FILE="tmp_last_captured_error.txt"
+
+# __TMP_ENV is used to store the user's env vars so that we can then re-apply them after having sourced
+# their config file. This allows us to make command line vars take precedence over what's in the config file.
+readonly __TMP_ENV="tmp.env"
```
palicena, line 111: As the comment says, this variable will temporarily hold the user's environment variables and reverse the order of priority between command-line vars and config_file vars.

Previously, the order of operations was this:
1. inherit user's environment variables + those specified via the command line
2. apply the config file variables, possibly stomping over those from #1

Now, the order is:
1. (same as previous -- inherit env from user and CLI)
2. store all current environment variables, i.e. those we got in step 1
3. apply the config file variables
4. re-apply the env variable stored in step 2

Note that there's no way to differentiate environment variables provided via the command line VS those that were already set in the shell before the script was called, so it's still not as "ideal" as having proper script options (i.e. `--my-var=something`), but it's one step closer.

cluster-install/install.sh
```diff
@@ -203,6 +207,17 @@ set_log_level() {
...
+ingest_config_string() {
+    local -r config_str="$1"
+    [ -z "$config_str" ] && fatal "no config string given"
+    local -r tmp_env="$__TMP_ENV"
+
+    echo "$config_str" > "$tmp_env"
+    # shellcheck disable=SC1090
+    source "$tmp_env"
+    rm -f "$tmp_env" &> /dev/null
+}
```
palicena, line 219: This is really the only place that new temporary env file is used. There's probably a way to do what this function is doing without creating a temporary file, but no matter what bash syntax I threw at the script, I couldn't really get it working. Decided not to waste too much time and just use what works for now.

cluster-install/install.sh
```diff
@@ -214,12 +229,33 @@ load_config_from_file_if_given() {
...
+        # variable's value (such as in b64 encoded values).
+        key="${line%%=*}"
+
+        # This does something similar, but this time removing the shortest match (one '#') of '*='
+        # (wildcard + equal sign) from the beginning ('#' as opposed to '%'), keeping only the value.
+        value="${line#*=}"
+
+        # Rebuild the env with the double quotes
+        previous_env_double_quoted+="$key=\"$value\"${__NEWLINE}"
+    done < <(env)
```
palicena, line 249: Getting the user's env was pretty easy (just call `env`), but unfortunately, that trimmed out any surrounding quotes from the values, causing issues when we re-applied them, so we have to re-quote the values ourselves.

cluster-install/install.sh
```diff
@@ -1004,10 +1040,9 @@ get_registry_credentials_from_pull_secret() {
     # 6. Get the first matching entry's `.value.auth`, which would be `B64_ENCODED_USERNAME_PASSWORD` here
     # 7. The `B64_ENCODED_USERNAME_PASSWORD` can then be passed to curl via
     # `-H 'Authorization: Basic B64_ENCODED_USERNAME_PASSWORD'` without having to decode it first.
-    local -r contents="$(kubectl get secret "$pull_secret" -n "$namespace" -o json 2> /dev/null)"
+    local -r contents="$(k8s_get_resource "secret/$pull_secret" "$namespace" "json")"
```
palicena, line 1043: Added a two new functions that are a lot of the changes in this PR:
- `k8s_get_resource`
- `k8s_resource_exists`
Both are pretty self-explanatory, but the main reason for their existence is to handle the separation of different errors (not found vs timeout/connection/etc.) we might encounter when trying to run these kubectl commands. 

A good example for why we need this is how we detect whether Trident is installed or not. Right now, any error will essentially lead to us detecting that "trident is not installed", even if the error was due to a timeout or a temporary connection blip. This causes the script to continue execution, assuming trident is not installed, which then leads to it reinstalling Trident and potentially stomping all over a user's Trident install.

You'll get to see the actual function def just a bit below, where I'll point out the different errors being handled.

cluster-install/install.sh
```diff
@@ -1183,6 +1218,53 @@ EOF
...
+    if echo "$captured_err" | grep -q "NotFound" &> /dev/null; then
+        logdebug "got NotFound error for resource '$resource', letting it through" >& 2
+    elif echo "$captured_err" | grep -q "error executing jsonpath"; then
+        logdebug "got jsonpath error for resource '$resource', letting it through" >& 2
+    elif [ -n "$captured_err" ]; then
+        logdebug "got unrecognized error for resource '$resource': $captured_err" >& 2
+        add_problem "$base_msg: $(echo "$captured_err" | tail -n 1)"
+    else
+        add_problem "$base_msg: unknown error"
+    fi
```
palicena, line 1254: This is where we handle the different errors. Both `NotFound` and `jsonpath` errors will return an exit code of 1 (indicating the resource wasn't found), while others add an actual problem, which will cause the script to exit and display the error returned by k8s.

For example:
```
* A failure occurred when checking if resource 'crd/volumesnapshotclasses.snapshot.storage.k8s.io' exists: Unable to connect to the server: dial tcp: some.bad.api.server.url: no such host
```

cluster-install/install.sh
```diff
@@ -1253,7 +1337,7 @@ wait_for_cr_state() {
     local counter=0
     local current_state=""
     while ((counter < max_checks)); do
-        current_state=$(kubectl get -n "$namespace" "$resource" -o jsonpath="{$path}" 2> /dev/null)
+        current_state=$(k8s_get_resource "$resource" "$namespace" jsonpath="{$path}")
```
palicena, line 1340: As mentioned, you'll be seeing a lot of these since I replaced pretty much all `kubectl get` calls.

cluster-install/install.sh
```diff
@@ -2450,13 +2533,19 @@ step_generate_and_apply_resource_limit_patches() {
     # itself does "support" resource limits, in the sense that they don't get cleared out when we set them.
 
     # Trident Operator
-    if ! trident_will_be_installed_or_modified; then
+    if [ -z "$_USER_ACCEPTED_TRIDENT_OPERATOR_UPGRADE" ]; then
+        logdebug "no trident upgrade operator was needed, skipping trident operator resource limits"
+    elif [ "$_USER_ACCEPTED_TRIDENT_OPERATOR_UPGRADE" != "true" ]; then
+        logdebug "user declined trident operator upgrade, skipping trident operator resource limits"
```
palicena, line 2539: This is where we check if the user declined the upgrade, and block applying resource limits if so.

cluster-install/install.sh
```diff
@@ -2450,13 +2533,19 @@ step_generate_and_apply_resource_limit_patches() {
...
+    elif ! trident_will_be_installed_or_modified; then
         logdebug "skipping trident operator resource limits"
     elif create_kubectl_patch_for_containers "deploy/trident-operator" "$trident_ns" "$patch_path" "$patch_value"; then
         patches_list_for_debugging+=("$_return_value")
         apply_kubectl_patch "$_return_value"
+    elif [ -n "$_return_error" ]; then
+        add_problem "Failed to create resource limit patch for the Trident Operator deployment: $_return_error"
     else
-        add_problem "Failed to create resource limit patch for the Trident Operator deployment (unknown error)"
+        add_problem "Failed to create resource limit patch for the Trident Operator deployment: unknown error"
```
palicena, line 2548: The `create_kubectl_patch_for_containers` function was calling `fatal` on some errors that were instead more appropriate as "problems" to be communicated to the user properly. These changes are to handle those errors.

cluster-install/install.sh
```diff
@@ -1355,11 +1439,14 @@ create_kubectl_patch_for_containers() {
         container_count=1
         logdebug "resource '$resource' in namespace '$namespace' not found, continuing anyway because this is a dry run"
     else
-        fatal "resource '$resource' in namespace '$namespace' was never created"
+        logdebug "resource '$resource' in namespace '$namespace' was never created"
+        _return_error="exceeded timeout waiting for '$resource' in namespace '$namespace' to be created"
+        return 1
```
palicena, line 1444: These errors were previously called as `fatal` errors, which wasn't really appropriate because this is just a resource failing to come up, not a flaw in the script logic. "Returning" them will allow the code calling the function to handle the error properly (in this case, adding it as a "problem", with a proper message to the user explaining the failure). I'll point it out in another comment.

cluster-install/install.sh
palicena, line 2548: As mentioned above, the `create_kubectl_patch_for_containers` function was calling `fatal` on some errors that were instead more appropriate as "problems" to be communicated to the user properly. These changes are to handle those errors.

cluster-install/install.sh
```diff
@@ -2617,16 +2713,26 @@ if trident_will_be_installed_or_modified; then
...
+                        step_generate_trident_operator_patch
+                        _USER_ACCEPTED_TRIDENT_OPERATOR_UPGRADE="true"
+                    fi
                 else
-                    loginfo "Trident will not be upgraded."
+                    _msg="You have chosen to use a version of Trident that is not supported with the current version"
+                    _msg+=" of Astra Control. This may result in some App Data Management operations not functioning"
+                    _msg+=" correctly or being blocked within Astra Control. It is highly recommended to upgrade"
+                    _msg+=" Trident to ensure compatibility and proper functionality."
+                    logwarn "$_msg"
```
palicena, line 2725: This is a new warning message for users declining Trident upgrades, as requested by Vijitha.

palicena:
Given sufficient approvals, I encourage any of the reviewers to merge the PR since I'll be on vacation next week.

unified-installer/astra-unified-installer.sh
```diff
@@ -188,6 +212,90 @@ get_configs() {
...
 }
 
+print_help() {
+    cat <<EOF
+Here is a list of available options and environment variables for configuring the Astra cluster installation script.
+Values which are required and have no default value are marked with '*'.
+
+Options:
+  --help/-h    Show this message and exit.
+  --required   Prints all required environment variables which do not have a default value.
```
nhammern, line 222: another way I've seen this done is to simply mark them as REQUIRED in the list we print out with --help. Have you considered that option?

palicena: Currently they're marked with `*`, but it's essentially the same thing with a different marker. My concern does remain that there's so many environment variables, it's hard to really tell at a glance what a user _needs_ to provide, even with a more obvious marker (such as REQUIRED).

Here's the changes I had in mind:
```
Usage: 
  ENV_VAR_1="value" ENV_VAR_2="value" ./astra-unified-installer.sh [options]

Options:
  --help/-h    Show this message and exit.

Required Environment Variables:
  KUBECONFIG                        Path to the KUBECONFIG for the cluster you'd like to manage. Required.
  ASTRA_API_TOKEN                   The Astra API token. Required, provided by the Astra Control UI.
  ASTRA_CONTROL_URL                 The Astra Control URL. Required, provided by the Astra Control UI.
  ASTRA_ACCOUNT_ID                  The Astra account ID. Required, provided by the Astra Control UI.

Optional Environment Variables:
  ----- Script Functionality
  CONFIG_FILE                        Path to a configuration file. Overrides environment variables specified via command line.
  DRY_RUN                            Skips applying generated resources when set to true. Default is false.
  SKIP_IMAGE_CHECK                   Skips checking if images exist when set to true. Default is false.
  SKIP_ASTRA_CHECK                   Skips all checks requiring a connection to Astra Control when set to true. Default is false.
  SKIP_TLS_VALIDATION                Skips TLS validation for all requests to Astra Control, including the Connector (unless CONNECTOR_SKIP_TLS_VALIDATION is set) when set to true. Default is false.
  DISABLE_PROMPTS                    Skips all prompts, answering 'yes' by default when set to true. Default is false.
  DO_NOT_MODIFY_EXISTING_TRIDENT     Prevents any and all modification to the existing Trident installation (if any). Required if DISABLE_PROMPTS is true, otherwise defaults to false.

  ----- General Configuration
  COMPONENTS                         Determines what will be installed/upgraded. Default is ALL_ASTRA_CONTROL.
  IMAGE_PULL_SECRET                  Image pull secret for the Docker registry.
  NAMESPACE                          Overrides EVERY resource's namespace (for fresh installs only, not upgrades).
  LABELS                             Labels to be added to the generated resources (disclaimer: does not apply labels to resources created by the operators).
  RESOURCE_LIMITS_PRESET             Resource limit preset to use. Overridden by RESOURCE_LIMITS_CUSTOM_CPU and RESOURCE_LIMITS_CUSTOM_MEMORY (disclaimer: resource limits not currently supported by Trident).
  RESOURCE_LIMITS_CUSTOM_CPU         Custom CPU resource limit. Plain number.
  RESOURCE_LIMITS_CUSTOM_MEMORY      Custom Memory resource limit (in 'Gi')'. Plain number (i.e. do not include the 'Gi').

  ----- Image Configuration
    Image configuration is separated into three distinct parts:
      - Registry ("private.registry.io")
      - Repository ("some/repository/path/trident")
      - Tag ("24.02")

    Given the following images :
      - Trident Operator:    private.registry.io/some/repository/path/trident-operator:24.02
      - Trident Autosupport: private.registry.io/some/repository/path/trident-autosupport:24.02
      - Trident:             private.registry.io/some/repository/path/trident:24.02
      - Trident ACP:         private.registry.io/some/repository/path/trident-acp:24.02
      - Connector Operator:  private.registry.io/some/repository/path/astra-connector-operator:202405211614-main

    An appropriate configuration might be:
      - IMAGE_REGISTRY="private.registry.io"
      - IMAGE_BASE_REPO="some/repository/path"
      - TRIDENT_IMAGE_TAG="24.02"
      - CONNECTOR_OPERATOR_IMAGE_TAG="202405211614-main"

  IMAGE_REGISTRY                     The default Docker registry to pull all images from. Overridden by values below.
  TRIDENT_OPERATOR_IMAGE_REGISTRY    The Docker registry to pull the Trident Operator image from.
  TRIDENT_AUTOSUPPORT_IMAGE_REGISTRY The Docker registry to pull the Trident Autosupport image from.
  TRIDENT_IMAGE_REGISTRY             The Docker registry to pull the Trident image from.
  CONNECTOR_OPERATOR_IMAGE_REGISTRY  The Docker registry to pull the Connector Operator image from.
  CONNECTOR_IMAGE_REGISTRY           The Docker registry to pull the Connector image from.
  NEPTUNE_IMAGE_REGISTRY             The Docker registry to pull the Neptune image from.
  TRIDENT_ACP_IMAGE_REGISTRY         The Docker registry to pull the Trident ACP image from.

  IMAGE_BASE_REPO                    The default base repository path (i.e. excluding the image name) for all images. Overridden by values below.
  TRIDENT_OPERATOR_IMAGE_REPO        The repository path for the Trident Operator image.
  TRIDENT_AUTOSUPPORT_IMAGE_REPO     The repository path for the Trident Autosupport image.
  TRIDENT_IMAGE_REPO                 The repository path for the Trident image.
  CONNECTOR_OPERATOR_IMAGE_REPO      The repository path for the Connector Operator image.
  CONNECTOR_IMAGE_REPO               The repository path for the Connector image.
  NEPTUNE_IMAGE_REPO                 The repository path for the Neptune image.
  TRIDENT_ACP_IMAGE_REPO             The repository path for the Trident ACP image.

  TRIDENT_IMAGE_TAG                  The tag for the Trident image. Overridden by other 'TRIDENT_' values below.
  TRIDENT_OPERATOR_IMAGE_TAG         The tag for the Trident Operator image.
  TRIDENT_AUTOSUPPORT_IMAGE_TAG      The tag for the Trident Autosupport image.
  TRIDENT_ACP_IMAGE_TAG              The tag for the Trident ACP image.
  CONNECTOR_OPERATOR_IMAGE_TAG       The tag for the Connector Operator image.

  ----- Connector Configuration
  ASTRA_CLOUD_ID                     The Astra cloud ID. If not provided,
  ASTRA_CLUSTER_ID                   The Astra cluster ID. Required, provided by the Astra Control UI.
  CONNECTOR_HOST_ALIAS_IP            Sets a host alias for the Astra Connector pod.
  CONNECTOR_SKIP_TLS_VALIDATION      (WARNING: Not for production use!) Skips TLS validation for the Connector's requests to Astra Control if set to true.
  CONNECTOR_AUTOSUPPORT_ENROLLED     Enrolls the Connector in autosupport if set to true. Default is false.

```

nhammern: ah, my bad. missed the first line printed above 🤦 

I like the new organization your proposed. Either works for me.

PR #337, opened 2024-05-26
ASTRACTL-32772: pull secret fix

palicena:
- Add help message and arg processing
- Add `IMAGE_PULL_SECRET` to `trident-operator` and/or `torc` during upgrades if they're missing from the list
- Add prompt/warning if `IMAGE_PULL_SECRET=""` and the user is using custom images

Tested these changes through 4 scenarios:
1. Greenfield (no initial connector/trident install)
2. Brownfield (no initial connector, but pre-existing trident at 23.07)
3. Brownfield, but DO_NOT_MODIFY_EXISTING_TRIDENT=true (same setup as 2nd test)
4. Greenfield at version 23.07 + modified pull secrets + rerun script with upgrade to verify that pull secrets are properly added

### DIFF
--- a/cluster-install/install.sh
+++ b/cluster-install/install.sh
@@ -2593,11 +2593,10 @@ step_generate_trident_operator_patch() {
         if echo "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" | grep -q "^${IMAGE_PULL_SECRET}$" &> /dev/null; then
             logdebug "image pull secret '$IMAGE_PULL_SECRET' already present in trident-operator"
         else
-            local -r secret_obj='{"name": "'"$IMAGE_PULL_SECRET"'"}'
             if [ -z "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" ]; then
-                patch_list+=',{"op":"replace","path":"/spec/template/spec/imagePullSecrets","value":['"$secret_obj"']}'
+                patch_list+=',{"op":"replace","path":"/spec/template/spec/imagePullSecrets","value":[{"name":"'"$IMAGE_PULL_SECRET"'"}]}'
             else
-                patch_list+=',{"op":"add","path":"/spec/template/spec/imagePullSecrets/-","value":'"$secret_obj"'}'
+                patch_list+=',{"op":"add","path":"/spec/template/spec/imagePullSecrets/-","value":{"name":"'"$IMAGE_PULL_SECRET"'"}}'
             fi
         fi
     fi
@@ -2618,6 +2617,7 @@ step_generate_torc_patch() {
     if [ -z "$torc_name" ]; then fatal "no trident orchestrator name was given"; fi
 
     logheader $__DEBUG "Generating TORC patch"
+
     local torc_patch_list=""
 
     # Trident
@@ -2643,9 +2643,8 @@ step_generate_torc_patch() {
         if echo "$_EXISTING_TORC_PULL_SECRETS" | grep -q "^${IMAGE_PULL_SECRET}$" &> /dev/null; then
             logdebug "image pull secret '$IMAGE_PULL_SECRET' already present in torc"
         else
-            local -r secret_obj='{"name": "'"$IMAGE_PULL_SECRET"'"}'
             if [ -z "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" ]; then
-                torc_patch_list+='{"op":"replace","path":"/spec/imagePullSecrets","value":"'"$IMAGE_PULL_SECRET"'"},'
+                torc_patch_list+='{"op":"replace","path":"/spec/imagePullSecrets","value":["'"$IMAGE_PULL_SECRET"'"]},'
             else
                 torc_patch_list+='{"op":"add","path":"/spec/imagePullSecrets/-","value":"'"$IMAGE_PULL_SECRET"'"},'
             fi
--- a/cluster-install/install.sh
+++ b/cluster-install/install.sh
@@ -17,11 +17,13 @@ _KUBERNETES_VERSION=""
 # existing Trident information (if any) has been collected.
 _TRIDENT_COLLECTION_STEP_CALLED="false"
 _EXISTING_TORC_NAME="" # TORC is short for TridentOrchestrator (works with kubectl too)
+_EXISTING_TORC_PULL_SECRETS="" # Newline-delimited values, e.g. 'secret1\nsecret2\nsecret3'
 _EXISTING_TRIDENT_NAMESPACE=""
 _EXISTING_TRIDENT_IMAGE=""
 _EXISTING_TRIDENT_ACP_ENABLED=""
 _EXISTING_TRIDENT_ACP_IMAGE=""
 _EXISTING_TRIDENT_OPERATOR_IMAGE=""
+_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS="" # Newline-delimited values, e.g. 'secret1\nsecret2\nsecret3'
 _USER_ACCEPTED_TRIDENT_OPERATOR_UPGRADE="" # "true" if yes, "false" if no, "" if the question was never asked
 
 # _PATCHES_ variables contain the k8s patches that will be applied after we've applied all CRs and kustomize resources.
@@ -62,7 +64,7 @@ readonly __COMPONENTS_VALID_VALUES=("$__COMPONENTS_ALL_ASTRA_CONTROL" "$__COMPON
 readonly __DEFAULT_DOCKER_HUB_IMAGE_REGISTRY="docker.io"
 readonly __DEFAULT_DOCKER_HUB_IMAGE_BASE_REPO="netapp"
 readonly __DEFAULT_ASTRA_IMAGE_REGISTRY="cr.astra.netapp.io"
-readonly __DEFAULT_IMAGE_TAG="$__RELEASE_VERSION"
+readonly __DEFAULT_ASTRA_IMAGE_BASE_REPO=""
 
 readonly __DEFAULT_TRIDENT_OPERATOR_IMAGE_NAME="trident-operator"
 readonly __DEFAULT_TRIDENT_AUTOSUPPORT_IMAGE_NAME="trident-autosupport"
@@ -115,10 +117,28 @@ readonly __NEWLINE=$'\n' # This is for readability
 #----------------------------------------------------------------------
 #-- Script config
 #----------------------------------------------------------------------
+process_args() {
+    for arg in "$@"; do
+        case $arg in
+            --help|-h)
+                print_help
+                exit 0
+                ;;
+            --required)
+                local msg="Listed below are all required environment variables which do not have a default value"
+                msg+=" (and so must be provided by the user):${__NEWLINE}"
+                echo "$msg"
+                print_help | grep '\*[A-Z]\+'
+                exit 0
+                ;;
+        esac
+    done
+}
+
 get_configs() {
     # ------------ SCRIPT BEHAVIOR ------------
     CONFIG_FILE="${CONFIG_FILE:-}" # Overrides environment variables specified via command line
-    DRY_RUN="${DRY_RUN:-"true"}" # Skips applying generated resources
+    DRY_RUN="${DRY_RUN:-"false"}" # Skips applying generated resources
     SKIP_IMAGE_CHECK="${SKIP_IMAGE_CHECK:-"false"}" # Skips checking whether images exist or not
     SKIP_ASTRA_CHECK="${SKIP_ASTRA_CHECK:-"false"}" # Skips AC URL, cloud ID, and cluster ID check
     # DISABLE_PROMPTS skips prompting the user when something notable is about to happen (such as a Trident Upgrade).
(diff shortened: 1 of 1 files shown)


cluster-install/install.sh
```diff
@@ -17,11 +17,13 @@ _KUBERNETES_VERSION=""
 # existing Trident information (if any) has been collected.
 _TRIDENT_COLLECTION_STEP_CALLED="false"
 _EXISTING_TORC_NAME="" # TORC is short for TridentOrchestrator (works with kubectl too)
+_EXISTING_TORC_PULL_SECRETS="" # Newline-delimited values, e.g. 'secret1\nsecret2\nsecret3'
 _EXISTING_TRIDENT_NAMESPACE=""
 _EXISTING_TRIDENT_IMAGE=""
 _EXISTING_TRIDENT_ACP_ENABLED=""
 _EXISTING_TRIDENT_ACP_IMAGE=""
 _EXISTING_TRIDENT_OPERATOR_IMAGE=""
+_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS="" # Newline-delimited values, e.g. 'secret1\nsecret2\nsecret3'
```
palicena, line 26: These two will hold the existing pull secrets, which we then use when crafting corresponding upgrade patches to determine whether we should add the current IMAGE_PULL_SECRET or not.

cluster-install/install.sh
```diff
@@ -62,7 +64,7 @@ readonly __COMPONENTS_VALID_VALUES=("$__COMPONENTS_ALL_ASTRA_CONTROL" "$__COMPON
 readonly __DEFAULT_DOCKER_HUB_IMAGE_REGISTRY="docker.io"
 readonly __DEFAULT_DOCKER_HUB_IMAGE_BASE_REPO="netapp"
 readonly __DEFAULT_ASTRA_IMAGE_REGISTRY="cr.astra.netapp.io"
-readonly __DEFAULT_IMAGE_TAG="$__RELEASE_VERSION"
+readonly __DEFAULT_ASTRA_IMAGE_BASE_REPO=""
```
palicena, line 67: Seems silly to have a default constant set to empty, but it's used in 2 places and I didn't want those to get unsynchronized in case we ever change this value.

cluster-install/install.sh
```diff
@@ -62,7 +64,7 @@ readonly __COMPONENTS_VALID_VALUES=("$__COMPONENTS_ALL_ASTRA_CONTROL" "$__COMPON
 readonly __DEFAULT_DOCKER_HUB_IMAGE_REGISTRY="docker.io"
 readonly __DEFAULT_DOCKER_HUB_IMAGE_BASE_REPO="netapp"
 readonly __DEFAULT_ASTRA_IMAGE_REGISTRY="cr.astra.netapp.io"
-readonly __DEFAULT_IMAGE_TAG="$__RELEASE_VERSION"
```
palicena, line 65: Dead code (was no longer being used).

cluster-install/install.sh
```diff
@@ -115,10 +117,28 @@ readonly __NEWLINE=$'\n' # This is for readability
...
+            --required)
+                local msg="Listed below are all required environment variables which do not have a default value"
+                msg+=" (and so must be provided by the user):${__NEWLINE}"
+                echo "$msg"
+                print_help | grep '\*[A-Z]\+'
+                exit 0
+                ;;
+        esac
+    done
+}
```
palicena, line 136: Added arg processing to support the help flag.

cluster-install/install.sh
```diff
@@ -115,10 +117,28 @@ readonly __NEWLINE=$'\n' # This is for readability
...
+
 get_configs() {
     # ------------ SCRIPT BEHAVIOR ------------
     CONFIG_FILE="${CONFIG_FILE:-}" # Overrides environment variables specified via command line
-    DRY_RUN="${DRY_RUN:-"true"}" # Skips applying generated resources
+    DRY_RUN="${DRY_RUN:-"false"}" # Skips applying generated resources
```
palicena, line 141: Making this default to false, as requested by PMs.

cluster-install/install.sh
```diff
@@ -198,6 +218,90 @@ get_configs() {
     CONNECTOR_AUTOSUPPORT_URL="${CONNECTOR_AUTOSUPPORT_URL:-$__PRODUCTION_AUTOSUPPORT_URL}"
 }
 
+print_help() {
```
palicena, line 221: Help!

cluster-install/install.sh
```diff
@@ -408,6 +523,27 @@ acp_is_enabled() {
     return 1
 }
 
+config_has_at_least_one_custom_registry_or_repo() {
```
palicena, line 526: This check is used in tandem with checking if the `IMAGE_PULL_SECRET` is empty; if both are true, we prompt the user for a pull secret (or warn them if prompts are disabled).

cluster-install/install.sh
```diff
@@ -1599,8 +1755,17 @@ step_check_config() {
...
-        add_to_config_builder "IMAGE_PULL_SECRET"
+    elif config_has_at_least_one_custom_registry_or_repo; then
+        local custom_reg_warning="We detected one or more custom registry or repo values"
+        custom_reg_warning+=", but no IMAGE_PULL_SECRET was specified. If any of your images are hosted in a private"
+        custom_reg_warning+=" registry, an image pull secret will need to be created and IMAGE_PULL_SECRET set."
+        if prompts_disabled; then
+            logwarn "$custom_reg_warning"
+        elif prompt_user_yes_no "$custom_reg_warning${__NEWLINE}Would you like to specify a pull secret now?"; then
+            prompt_user "IMAGE_PULL_SECRET" "Enter a value for IMAGE_PULL_SECRET: "
+        fi
```
palicena, line 1766: This is where we prompt/warn the user for an IMAGE_PULL_SECRET if it's empty and the user specified one or more custom images.

cluster-install/install.sh
```diff
@@ -1670,12 +1835,18 @@ step_check_k8s_version_in_range() {
 
     logheader $__DEBUG "Checking if Kubernetes version is within range ($minimum <=> $maximum)..."
 
-    local current; current="$(kubectl version -o json | jq -r '.serverVersion.gitVersion')"
+    local current; current="$(kubectl version -o json 2> "$__ERR_FILE" | jq -r '.serverVersion.gitVersion' 2> /dev/null)"
+    local -r captured_err="$(get_captured_err)"
+    if [ -z "$current" ] || [ "$current" == null ] || [ -n "$captured_err" ]; then
+        add_problem "Failed to get your cluster's Kubernetes version: $captured_err"
+        return 1
+    fi
```
palicena, line 1843: Adding actually useful error message when we fail to get the version (due to a connection error or whatever).

cluster-install/install.sh
```diff
@@ -1688,12 +1859,19 @@ step_check_k8s_permissions() {
...
+    if [ "$has_permission" == "yes" ]; then
         logdebug "k8s permissions: OK"
+        return 0
+    elif [ "$has_permission" != "yes" ]; then
+        add_problem "Kubernetes user does not have admin privileges"
+    elif [ -n "$captured_err" ]; then
+        add_problem "Failed to check if Kubernetes user has admin privilege: $captured_err"
     else
-        add_problem "k8s permissions: user does not have admin privileges" "Kubernetes user does not have admin privileges"
+        add_problem "Failed to check if Kubernetes user has admin privilege: unknown error"
```
palicena, line 1872: Same here, just better error messaging.

cluster-install/install.sh
```diff
@@ -1708,22 +1886,48 @@ step_check_volumesnapshotclasses() {
...
+    # Best effort guess at which registry to put in the create command
+    local registry="<REGISTRY>"
+    if components_include_connector; then
+        registry="$CONNECTOR_IMAGE_REGISTRY"
+    # Skip if Trident's image is still set to our default of 'docker.io/netapp', which is public
+    elif components_include_trident && ! str_contains_at_least_one "$(get_config_trident_image)" "docker.io/netapp" ; then
+        registry="$TRIDENT_IMAGE_REGISTRY"
+    elif components_include_acp; then
+        registry="$TRIDENT_ACP_IMAGE_REGISTRY"
+    fi
```
palicena, line 1905: A bit further below, I add generate a `kubectl create secret` command if we can't find the pull secret in the given NAMESPACE. This big `if` block is used to determine which registry to prefill in the command, if any.

cluster-install/install.sh
```diff
@@ -1708,22 +1886,48 @@ step_check_volumesnapshotclasses() {
...
+    local create_secret_cmd="kubectl create secret docker-registry '$pull_secret' -n '$namespace'"
+    create_secret_cmd+=" --docker-server='$registry' --docker-username='<USERNAME>' --docker-password='<PASSWORD>'"
+
+    if ! k8s_resource_exists "namespace/$namespace"; then
+        if [ -n "$pull_secret" ]; then
+            err_msg="The specified NAMESPACE '$namespace' does not exist on the cluster, but IMAGE_PULL_SECRET is set."
+            err_msg+=" Please create the namespace and secret:"
+            err_msg+="${__NEWLINE}    - kubectl create namespace '$namespace'"
+            err_msg+="${__NEWLINE}    - $create_secret_cmd"
+            add_problem "namespace '$namespace': not found but IMAGE_PULL_SECRET is set" "$err_msg"
```
palicena, line 1916: Here we add the generated command to the error/problem message. Can't create the secret ourselves unfortunately, unless we decide to prompt them for their credentials, but that made for too many prompts/work so I left it as an error message.

cluster-install/install.sh
```diff
@@ -1708,22 +1886,48 @@ step_check_volumesnapshotclasses() {
...
         else
-            local -r err_msg="User chose not to create the namespace. Please create the namespace and/or run the script again."
+            err_msg="User chose not to create the namespace. Please create the namespace and/or run the script again."
             add_problem "User chose not to create the namespace. Exiting." "$err_msg"
             exit_if_problems
         fi
+    elif [ -n "$pull_secret" ] && ! k8s_resource_exists "secret/$pull_secret" "$namespace"; then
+        err_msg="The specified IMAGE_PULL_SECRET '$pull_secret' does not exist in namespace '$namespace'."
+        err_msg+=" Please create it:${__NEWLINE}    - $create_secret_cmd"
+        add_problem "pull secret '$pull_secret' not found in namespace '$namespace'" "$err_msg"
```
palicena, line 1930: This is specifically if the namespace exists, but the pull secret doesn't. We also print the generated pull secret command (but not the namespace one).

cluster-install/install.sh
```diff
@@ -2067,6 +2271,27 @@ EOF
     logdebug "$kustomization_file: added namespace transformer ($transformer_file_name)"
 }
 
+step_kustomize_global_pull_secret_if_needed() {
```
palicena, line 2274: Turns out we weren't adding the IMAGE_PULL_SECRET to fresh installs of the Trident or Connector operators. Didn't catch it since those are open source, but would've been a problem for dark sites.

cluster-install/install.sh
```diff
@@ -2262,6 +2487,15 @@ step_collect_existing_trident_info() {
...
     fi
 
+    # TORC imagePullSecrets
+    local -r torc_pull_secrets="$(echo "$torc_json" | jq -r '.spec.imagePullSecrets | join("\n")' 2> /dev/null)"
+    if [ -n "$torc_pull_secrets" ] && [ "$torc_pull_secrets" != "null" ]; then
+        logdebug "torc pull secrets:${__NEWLINE}$torc_pull_secrets"
+        _EXISTING_TORC_PULL_SECRETS="$torc_pull_secrets"
+    else
+        logdebug "torc pull secrets: not found"
+    fi
```
palicena, line 2497: Get the existing TORC pull secrets so we can later fiddle with it/determine whether we need to add the current IMAGE_PULL_SECRET to when generating upgrade patches. Probably could've made `_EXISTING_TORC_PULL_SECRETS` a boolean like "existing torc already has pull secret" or something, but I didn't think of that and I don't have time to fix it >.>

cluster-install/install.sh
```diff
@@ -2279,6 +2513,14 @@ step_collect_existing_trident_info() {
     else
         logdebug "trident operator: not found"
     fi
+    
+    local -r op_pull_secrets="$(echo "$trident_operator_json" | jq -r '[.spec.template.spec.imagePullSecrets.[] | .name] | join("\n")' 2> /dev/null)"
+    if [ -n "$op_pull_secrets" ] && [ "$op_pull_secrets" != "null" ]; then
+        logdebug "trident operator pull secrets:${__NEWLINE}$op_pull_secrets"
+        _EXISTING_TRIDENT_OPERATOR_PULL_SECRETS="$op_pull_secrets"
+    else
+        logdebug "trident operator pull secrets: not found"
```
palicena, line 2522: We do the same as the TORC pull secrets, but for the trident-operator.

cluster-install/install.sh
```diff
@@ -2343,8 +2585,28 @@ step_generate_trident_operator_patch() {
...
+            logdebug "image pull secret '$IMAGE_PULL_SECRET' already present in trident-operator"
+        else
+            local -r secret_obj='{"name": "'"$IMAGE_PULL_SECRET"'"}'
+            if [ -z "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" ]; then
+                patch_list+=',{"op":"replace","path":"/spec/template/spec/imagePullSecrets","value":['"$secret_obj"']}'
+            else
+                patch_list+=',{"op":"add","path":"/spec/template/spec/imagePullSecrets/-","value":'"$secret_obj"'}'
+            fi
+        fi
+    fi
```
palicena, line 2603: This is where we actually generate the trident operator patch that will add the IMAGE_PULL_SECRET, if needed.

cluster-install/install.sh
```diff
@@ -2376,7 +2638,21 @@ step_generate_torc_patch() {
...
+            logdebug "image pull secret '$IMAGE_PULL_SECRET' already present in torc"
+        else
+            local -r secret_obj='{"name": "'"$IMAGE_PULL_SECRET"'"}'
+            if [ -z "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" ]; then
+                torc_patch_list+='{"op":"replace","path":"/spec/imagePullSecrets","value":"'"$IMAGE_PULL_SECRET"'"},'
+            else
+                torc_patch_list+='{"op":"add","path":"/spec/imagePullSecrets/-","value":"'"$IMAGE_PULL_SECRET"'"},'
+            fi
+        fi
+    fi
```
palicena, line 2653: Same here, but for the TORC -- generate a patch to add the IMAGE_PULL_SECRET.

cluster-install/install.sh
```diff
@@ -2432,17 +2708,31 @@ step_apply_resources() {
...
-        output="$(kubectl apply -k "$operators_dir" 2>&1)"
+        output="$(kubectl apply -k "$operators_dir" 2> "$__ERR_FILE")"
+        captured_err="$(get_captured_err)"
+        if echo "$captured_err" | grep -q "Warning:"; then
+            logdebug "captured warning when applying kustomize resources:${__NEWLINE}$captured_err"
+        elif echo "$captured_err" | grep -q "no objects passed to apply"; then
+            logdebug "no kustomize resources to apply, skipping"
+        elif [ -z "$output" ] || [ -n "$captured_err" ]; then
+            add_problem "Failed to apply kustomize resources: $captured_err"
+        fi
```
palicena, line 2721: Better error handling.

cluster-install/install.sh
```diff
@@ -2432,17 +2708,31 @@ step_apply_resources() {
...
     # Apply CRs (if we have any)
     if [ -f "$crs_file_path" ]; then
         logdebug "apply CRs"
         if ! is_dry_run; then
-            output="$(kubectl apply -f "$crs_file_path")"
+            output="$(kubectl apply -f "$crs_file_path" 2> "$__ERR_FILE")"
+            captured_err="$(get_captured_err)"
+            if [ -z "$output" ] || [ -n "$captured_err" ]; then
+                add_problem "Failed to apply CRs: $captured_err"
+            fi
```
palicena, line 2735: Better error handling here too.

palicena:
Given sufficient approvals, I encourage any of the reviewers to merge the PR since I'll be on vacation next week.

cluster-install/install.sh
```diff
@@ -2641,8 +2932,9 @@ step_cleanup_tmp_files() {
 #======================================================================
 #== Main
 #======================================================================
+process_args "$@"
 set_log_level
-logln $__INFO "====== Astra Cluster Installer ${__RELEASE_VERSION} ======"
+logln $__INFO "====== Astra Unified Installer ${__RELEASE_VERSION} ======"
```
palicena, line 2937: **Astra Unified Installer** has been chosen as the official name of the script by Vijitha/PMs. Related card to deal with the folder name (which requires a change in both this repo and the polaris repo): https://jira.ngage.netapp.com/browse/ASTRACTL-33688

jharrod:
I think this makes the assumption the same pull secret will work for Trident images and AC images. I'm not sure if that is safe to assume or not.

zachatnetapp:
> I think this makes the assumption the same pull secret will work for Trident images and AC images. I'm not sure if that is safe to assume or not.

Right, I think also this might not be applying the secret to the right spot. Wouldn't we want the secret in trident-controller, not trident operator? Or maybe we want both. But the problem I (and hemantha) ran into was, because trident-controller didn't have the necessary secret to access jfrog (where we have trident-acp coming from), that image wasn't pulling.

cluster-install/install.sh
zachatnetapp: I think the value is not quite right. When we do the replace, we do want to create an array like you have, but the value in the image pull secrets array looks more like `{"name": "secret-name"}` instead of just "secret-name". I'll make this change here before I merge this PR

cluster-install/install.sh
zachatnetapp: Same comment about the values. I'll make the change on your behalf :D

zachatnetapp:
> I think this makes the assumption the same pull secret will work for Trident images and AC images. I'm not sure if that is safe to assume or not.

I think this assumption is safe for now.

cluster-install/install.sh
zachatnetapp: I merged a fix here, the add and replace operations had a small issue. Should be working now.

PR #338, opened 2024-05-28
adding label under cr in appriopriate spot

oscarr-ntap:
Labels are also needed under the astra spec in the CR so that operator knows to create different resource with appropriate labels. This adds the labels there without default label "app.kubernetes.io/created-by: astra-cluster-install-script"


cluster-install/install.sh
```diff
@@ -2080,14 +2097,14 @@ apiVersion: astra.netapp.io/v1
...
-  namespace: "${connector_namespace}"${labels_field_and_content}
+  namespace: "${connector_namespace}"${labels_field_and_content_with_default}
 spec:
   astra:
     accountId: ${account_id}
     tokenRef: astra-api-token
     cloudId: ${cloud_id}
     clusterId: ${cluster_id}
-    skipTLSValidation: ${skip_tls_validation}  # Should be set to false in production environments
+    skipTLSValidation: ${skip_tls_validation}  # Should be set to false in production environments${labels_field_and_content}
```
josecrosa, line 2107: Did you mean to add this after the comment?

oscarr-ntap: yeah it's the way to ensure it's there after the skipTLSValidation.. also down above and other places.. might be a better way though

oscarr-ntap: ChatGPT4 - "The approach you're using to conditionally include ${labels_field_and_content} in your bash script is a common and effective one. By appending the variable directly after the namespace field, it will only take up space if it's not empty. This is a good way to handle optional content in bash scripts."

josecrosa: Well, if our AI overlords says so... 👍

josecrosa (approved):

cluster-install/install.sh
jharrod: I think it works because 'labels_field_and_content' starts with a newline when set. I agree this looks odd but is probably ok

jharrod (approved):

PR #343, opened 2024-05-31
Astractl 33663 set resource defaults and make them mutable via CR

oscarr-ntap:
Changing the defaults of Astra Connector and Neptune to be the default "small" established in
https://jira.ngage.netapp.com/browse/ASTRACTL-32992

The resource size can also now be passed in via the CR.


josecrosa (approved):

app/deployer/connector/astra_connect_natless.go
```diff
@@ -55,6 +55,22 @@ func (d *AstraConnectDeployer) GetDeploymentObjects(m *v1.AstraConnector, ctx co
 	connectorImage = fmt.Sprintf("%s/astra-connector:%s", imageRegistry, containerImage)
 	log.Info("Using AstraConnector image", "image", connectorImage)
 
+	var connectResourceSize corev1.ResourceRequirements
+	if m.Spec.AstraConnect.ResourceRequirements.Limits == nil && m.Spec.AstraConnect.ResourceRequirements.Requests == nil {
```
alexkarlis, line 59: Is it possible that `m.Spec.AstraConnect.ResourceRequirements.Limits != nil && m.Spec.AstraConnect.ResourceRequirements.Requests == nil`? Or vice versa?

alexkarlis: If so, then we'll need to break this if statement into 2

oscarr-ntap: If there are both nil that means the customer did not specify anything in the inpu so setting default, if one or the other is not nil that means the customer specified it and i am just setting to exactly the ResourceRequirements they put in. Should i handle each requirement separately you think?

alexkarlis (approved):

PR #345, opened 2024-05-31
ASTRACTL-33797: handles manifest image registry for release branch

colep-ntap:
We were previously assuming that all images were pushed to jfrog when creating the image manifest file from cicd. This adds a handler for the case that we are in a release branch and those images will be in the ACS registry instead. Currently we are not handling the alternate tags that these images might have, that will be included in this script when we have an idea of how those images will be tagged for the upcoming release.


jharrod (approved):

common/constants.go
```diff
@@ -13,6 +13,7 @@ import (
 
 const (
 	DefaultImageRegistry = "netappdownloads.jfrog.io/docker-astra-control-staging/arch30/neptune"
+	AstraImageRegistry   = "cr.astra.netapp.io"
```
jharrod, line 16: @mgunaseg plz confirm there aren't any subdirectories, e.g. "neptune" in the ACS repo?
otherwise we might need this to be "cr.astra.netapp.io/neptune"

PR #347, opened 2024-06-01
Astractl-33663 Scale Memory on Neptune based on estimated snapshots and backups

oscarr-ntap:
We will set the default of 2GB of memory for users who anticipate having 10000 snapshots/backups or less retained at a single time. And increase 1GB for every 5 thousand after that. More info here.
https://jira.ngage.netapp.com/browse/ASTRACTL-33663

Example out put with answer below 
```
Do you anticipate having more than 10,000 snapshots and backups existing at the same time at any point?  (yes/no): yes
Please estimate the maximum number of snapshots and backups you expect to have existing simultaneously within this cluster? (enter number value): 24000
Memory limit set to: 5 GB
```

tested with values under 10,000 and with values above 10,000


unified-installer/astra-unified-installer.sh
```diff
@@ -2036,28 +1849,22 @@ step_generate_astra_connector_yaml() {
...
-metadata:
-  name: operator-controller-manager
-  namespace: $original_acop_namespace
-spec:
-  template:
-    spec:
-      containers:
-        - name: manager
-          env:
-            - name: ACOP_FEATUREFLAGS_NATLESS
```
oscarr-ntap, line 2057: This is no longer needed. Feature flag has been deleted from code

alexkarlis (approved):

PR #348, opened 2024-06-03
astractl-33663: scale memory limits based on snpahot/backup count

jharrod:
Fixes a "division by zero" ark error. The root cause still confuses me but I was unable to reproduce it with these changes.

I also adjusted the math and added a comment of how we were calculating.

### DIFF
--- a/unified-installer/astra-unified-installer.sh
+++ b/unified-installer/astra-unified-installer.sh
@@ -1855,7 +1855,7 @@ step_generate_astra_connector_yaml() {
     memory_limit=2
     if [  "$DISABLE_PROMPTS" != "true" ]; then
       if prompt_user_yes_no "Do you anticipate having more than 10,000 snapshots and backups existing at the same time at any point? "; then
-          snapshot_count="" # Default snapshot count
+          local snapshot_count="" # Default snapshot count
           prompt_user_number_greater_than_zero snapshot_count "Please estimate the maximum number of snapshots and backups you expect to have existing simultaneously within this cluster? (enter number value): "
           # Calculate estimated_memory and round up to the nearest integer
 
--- a/unified-installer/astra-unified-installer.sh
+++ b/unified-installer/astra-unified-installer.sh
@@ -609,7 +609,7 @@ prompt_user_yes_no() {
     local prompt=$1
     local result
     if [ -z "$prompt" ]; then fatal "no prompt message was given"; fi
-    if [ "$DISABLE_PROMPTS" == "true" ]; then return 0; fi
+    if [ "$DISABLE_PROMPTS" == "true" ]; then return 1; fi
 
     echo
     while true; do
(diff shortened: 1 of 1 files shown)


unified-installer/astra-unified-installer.sh
```diff
@@ -1851,11 +1851,20 @@ step_generate_astra_connector_yaml() {
 
     # Default memory limit
     memory_limit=2
-    snapshot_count=""
     if prompt_user_yes_no "Do you anticipate having more than 10,000 snapshots and backups existing at the same time at any point? "; then
+        snapshot_count="10000" # Default snapshot count
         prompt_user_number_greater_than_zero snapshot_count "Please estimate the maximum number of snapshots and backups you expect to have existing simultaneously within this cluster? (enter number value): "
         # Calculate estimated_memory and round up to the nearest integer
-        estimated_memory=$(echo "$snapshot_count 5000" | awk '{printf("%d\n", ($1/$2)+0.6)}')
```
jharrod, line 1858: Double check my change here. I wasn't sure where the "0.6" was coming from

palicena: Sounds like a way to force rounding up when the default behavior is to round to the closest integer (up or down)? Or something like that.

unified-installer/astra-unified-installer.sh
```diff
@@ -1851,11 +1851,20 @@ step_generate_astra_connector_yaml() {
 
     # Default memory limit
     memory_limit=2
-    snapshot_count=""
```
palicena, line 1854: Just noticed, these are global since they're not declared as local -- any chance we could change them to local as part of this PR?

palicena (approved):

oscarr-ntap (approved):

unified-installer/astra-unified-installer.sh
jharrod: Put this back to how it was after talking with Oscar, but removed the 0.6 thing. It was mostly ok as is.

unified-installer/astra-unified-installer.sh
```diff
@@ -1851,21 +1853,25 @@ step_generate_astra_connector_yaml() {
...
-    if prompt_user_yes_no "Do you anticipate having more than 10,000 snapshots and backups existing at the same time at any point? "; then
-        prompt_user_number_greater_than_zero snapshot_count "Please estimate the maximum number of snapshots and backups you expect to have existing simultaneously within this cluster? (enter number value): "
-        # Calculate estimated_memory and round up to the nearest integer
-        estimated_memory=$(echo "$snapshot_count 5000" | awk '{printf("%d\n", ($1/$2)+0.6)}')
-
-        # If estimated_memory is greater than memory_limit, set memory_limit to estimated_memory
-        if (( estimated_memory > memory_limit )); then
-            memory_limit=$estimated_memory
-        fi
+    if [  "$DISABLE_PROMPTS" != "true" ]; then
```
jharrod, line 1856: This fixes the div by 0. 
'prompt_user_yes_no' below assumes the answer is always 'yes' when prompts are disabled.

unified-installer/astra-unified-installer.sh
oscarr-ntap: thanks for that, teach me the bash ways!

PR #349, opened 2024-06-04
ASTRACTL-32772: help message, prioritize CLI vars over config

palicena:
- Add the help message (`--help / -h`)
- Prioritize CLI env vars over config file
- Various fixes (see comments)


unified-installer/astra-unified-installer.sh
```diff
@@ -41,7 +41,7 @@ _PROCESSED_LABELS=""
 readonly __RELEASE_VERSION="24.02"
 readonly __TRIDENT_VERSION="${__TRIDENT_VERSION_OVERRIDE:-"$__RELEASE_VERSION"}"
 
-readonly -a __REQUIRED_TOOLS=("git" "jq" "kubectl" "curl" "grep" "sort" "uniq" "find" "base64" "wc" "awk")
+readonly -a __REQUIRED_TOOLS=("git" "jq" "kubectl" "curl" "grep" "sort" "uniq" "find" "base64" "wc" "awk" "tail" "head")
```
palicena, line 44: `tail` and `head` have been used since day 1 but were missing from the list.

unified-installer/astra-unified-installer.sh
```diff
@@ -99,15 +99,38 @@ readonly __FATAL=50
 #     captured_stdout="$(curl -sS https://bad-url.com 2> "$_ERR_FILE")"
 #     captured_stderr="$(get_captured_err)"
 readonly __ERR_FILE="tmp_last_captured_error.txt"
+
+# __TMP_ENV is used to store the user's env vars so that we can then re-apply them after having sourced
+# their config file. This allows us to make command line vars take precedence over what's in the config file.
+readonly __TMP_ENV="tmp.env"
```
palicena, line 105: As the comment mentions, this is used to temporarily hold the user's environment variables (not from a config file, but the ones inherited by the script from the parent shell). The order of operations are roughly:

1. Store inherited environment
2. Apply user's config file
3. Apply stored inherited environment

This allows us to reverse the priority of the variables, making CLI args take precedence over the config file. This is something that was requested by PMs, though the only reason it hadn't been done in the first place regardless was the challenge of doing it in bash being surprisingly not-so-easy to solve, as you'll see later (I'll point out the relevant code in a comment).

unified-installer/astra-unified-installer.sh
```diff
@@ -99,15 +99,38 @@ readonly __FATAL=50
...
+            --required)
+                local msg="Listed below are all required environment variables which do not have a default value"
+                msg+=" (and so must be provided by the user):${__NEWLINE}"
+                echo "$msg"
+                print_help | grep '\*[A-Z]\+'
+                exit 0
+                ;;
+        esac
+    done
+}
```
palicena, line 128: This is just a typical arg processing function. Right now, there's only `--help/-h` and `--required`. About the latter flag -- I thought that, with the very long list of variables available, having a flag that outputs only the variables that are absolutely required (i.e. we don't provide a default for) might be useful. If you think it's fluff and not very useful though, I could be convinced.

unified-installer/astra-unified-installer.sh
```diff
@@ -99,15 +99,38 @@ readonly __FATAL=50
...
+
 get_configs() {
     # ------------ SCRIPT BEHAVIOR ------------
     CONFIG_FILE="${CONFIG_FILE:-}" # Overrides environment variables specified via command line
-    DRY_RUN="${DRY_RUN:-"true"}" # Skips applying generated resources
+    DRY_RUN="${DRY_RUN:-"false"}" # Skips applying generated resources
```
palicena, line 133: PM request to default this to false, though it's been mentioned a few times by a few of you as well.

unified-installer/astra-unified-installer.sh
```diff
@@ -197,6 +302,17 @@ set_log_level() {
...
+ingest_config_string() {
+    local -r config_str="$1"
+    [ -z "$config_str" ] && fatal "no config string given"
+    local -r tmp_env="$__TMP_ENV"
+
+    echo "$config_str" > "$tmp_env"
+    # shellcheck disable=SC1090
+    source "$tmp_env"
+    rm -f "$tmp_env" &> /dev/null
+}
```
palicena, line 314: This function and the changes below are where we reverse the priority of CLI vars VS config vars. It's commented pretty heavily since it makes use of weird-looking bash magic.

unified-installer/astra-unified-installer.sh
```diff
@@ -690,7 +838,7 @@ join_rpath() {
     for (( i = 0; i < ${#args[@]}; i+=1 )); do
         args[i]="${args[i]#/}" # Remove starting slash if present
         args[i]="${args[i]%/}" # Remove trailing slash if present
-        if [ "$i" -eq 0 ]; then joined="${args[i]}"
+        if [ -z "$joined" ]; then joined="${args[i]}"
```
palicena, line 841: Fixes an issue for when the first arg(s) are empty values. 

Given args:
```
join_rpath "" "astra-connector"
```
You'd get:
```
"/astra-connector"
```
Now you'll instead get:
```
"astra-connector"
```
The extra slash at the beginning was breaking some image tags depending on the values provided.

unified-installer/astra-unified-installer.sh
```diff
@@ -1431,17 +1580,13 @@ step_check_config() {
...
-            if [ -z "${_PROCESSED_LABELS}" ]; then
-                add_problem "label processing: failed" "The given LABELS could not be parsed."
-            fi
+        _PROCESSED_LABELS+="$(process_labels_to_yaml "${LABELS}" "$label_indent")"
+        if [ -z "${_PROCESSED_LABELS}" ]; then
+            add_problem "label processing: failed" "The given LABELS could not be parsed."
         fi
-        add_to_config_builder "LABELS"
+    fi
+    add_to_config_builder "LABELS"
```
palicena, line 1589: Hard to see with how github aggressively hides contextual code surrounding changes, but there was some duplication here that could be easily resolved, so I did.

unified-installer/astra-unified-installer.sh
```diff
@@ -2054,6 +2198,7 @@ step_collect_existing_trident_info() {
 step_existing_trident_flags_compatibility_check() {
     [ "$COMPONENTS" == "$__COMPONENTS_ALL_ASTRA_CONTROL" ] && return 0
     ! existing_trident_needs_modifications && return 0
+    existing_trident_can_be_modified && return 0
```
palicena, line 2201: This check was missing from the function, making it think the DO_NOT_MODIFY_EXISTING_TRIDENT flag was always set to true.

unified-installer/astra-unified-installer.sh
```diff
@@ -2162,7 +2307,7 @@ step_add_labels_to_kustomization() {
     logheader $__DEBUG "Adding labels to kustomization and crs file"
 
     local -r content="labels:${__NEWLINE}- pairs:${__NEWLINE}${processed_labels}"
-    insert_into_file_after_pattern "${kustomization_file}" "kind:" "${content}"
+    insert_into_file_after_pattern "${kustomization_file}" "kind: Kustomization" "${content}"
```
palicena, line 2310: Some in-line kustomization patches have the `kind:` field, so I'm making this pattern more specific to avoid inserting the content in the wrong place (i.e. in the middle of a patch definition).

unified-installer/astra-unified-installer.sh
```diff
@@ -2202,16 +2347,29 @@ step_apply_resources() {
...
-        output="$(kubectl apply -k "$operators_dir" 2>&1)"
+        output="$(kubectl apply -k "$operators_dir" 2> "$__ERR_FILE")"
+        captured_err="$(get_captured_err)"
+        if echo "$captured_err" | grep -q "Warning:"; then
+            logdebug "captured warning when applying kustomize resources:${__NEWLINE}$captured_err"
+        elif echo "$captured_err" | grep -q "no objects passed to apply"; then
+            logdebug "no kustomize resources to apply, skipping"
+        elif [ -z "$output" ] || [ -n "$captured_err" ]; then
+            add_problem "Failed to apply kustomize resources: $captured_err"
+        fi
```
palicena, line 2358: Better error checking and logging.

unified-installer/astra-unified-installer.sh
```diff
@@ -188,6 +209,90 @@ get_configs() {
...
+  TRIDENT_AUTOSUPPORT_IMAGE_TAG      The tag for the Trident Autosupport image.
+  TRIDENT_ACP_IMAGE_TAG              The tag for the Trident ACP image.
+  CONNECTOR_OPERATOR_IMAGE_TAG       The tag for the Connector Operator image.
+
+  ----- Connector Configuration
+  *ASTRA_CONTROL_URL                 The Astra Control URL. Required, provided by the Astra Control UI.
+  *ASTRA_API_TOKEN                   The Astra API token. Required, provided by the Astra Control UI.
+  *ASTRA_ACCOUNT_ID                  The Astra account ID. Required, provided by the Astra Control UI.
+  *ASTRA_CLOUD_ID                    The Astra cloud ID. Required, provided by the Astra Control UI.
+  *ASTRA_CLUSTER_ID                  The Astra cluster ID. Required, provided by the Astra Control UI.
```
josecrosa, line 289: I thought we weren’t making these required after all?

palicena: You're right, I forgot we were doing that. I'll change the message.

unified-installer/astra-unified-installer.sh
josecrosa, line 105: Can we use a randomly generated name or something more unique like `install_script_env_file_helper.env`? I could see a tester or customer already having used “tmp.env” and us possibly overwriting it.

palicena: Good point, I'll add some generated value to the name.

unified-installer/astra-unified-installer.sh
jharrod: Why not put the required args in the help output? I think that's where a customer would expect to find it. Most people assume there is a -h or --help on CLI programs but would have to remember/read the docs to learn about the --required arg.

jharrod: "Listed below are all required environment variables which do not have a default value" maybe makes more sense to just say "Listed below are all required environment variables"

unified-installer/astra-unified-installer.sh
```diff
@@ -188,6 +209,90 @@ get_configs() {
...
+print_help() {
+    cat <<EOF
+Here is a list of available options and environment variables for configuring the Astra cluster installation script.
+Values which are required and have no default value are marked with '*'.
+
+Options:
+  --help/-h    Show this message and exit.
+  --required   Prints all required environment variables which do not have a default value.
+
+Environment Variables:
```
jharrod, line 221: Might make sense to split these up into required vars and optional vars

unified-installer/astra-unified-installer.sh
palicena: The required args are marked with `*` in the help output, but that gets somewhat lost in the sea of variables available, which is why the `--required` exists. We could re-order the help output to list the required variables first, but I thought it would be better to keep everything categorized properly. Maybe not? 

As for the wording, yeah, yours is better. I got a bit lost in the technicality initially since a lot of variables that have default _are_ required for the script to work, while some work fine even if empty, but from a user's perspective, they only care about what's required for them to input, not the technical behind-the-scenes meaning of "required". I'll change the message for it.

unified-installer/astra-unified-installer.sh
```diff
@@ -188,6 +209,90 @@ get_configs() {
...
+  SKIP_ASTRA_CHECK                   Skips all checks requiring a connection to Astra Control when set to true. Default is false.
+  SKIP_TLS_VALIDATION                Skips TLS validation for all requests to Astra Control, including the Connector (unless CONNECTOR_SKIP_TLS_VALIDATION is set) when set to true. Default is false.
+  DISABLE_PROMPTS                    Skips all prompts, answering 'yes' by default when set to true. Default is false.
+  DO_NOT_MODIFY_EXISTING_TRIDENT     Prevents any and all modification to the existing Trident installation (if any). Required if DISABLE_PROMPTS is true, otherwise defaults to false.
+
+  ----- General Configuration
+  *KUBECONFIG                        Path to the KUBECONFIG for the cluster you'd like to manage. Required.
+  COMPONENTS                         Determines what will be installed/upgraded. Default is ALL_ASTRA_CONTROL.
+  IMAGE_PULL_SECRET                  Image pull secret for the Docker registry.
+  NAMESPACE                          Overrides EVERY resource's namespace (for fresh installs only, not upgrades).
```
jharrod, line 235: What happens if NAMESPACE is set on an upgrade?

unified-installer/astra-unified-installer.sh
```diff
@@ -208,16 +324,38 @@ load_config_from_file_if_given() {
         add_problem "CONFIG_FILE '$config_file' does not exist" "Given CONFIG_FILE '$config_file' does not exist"
         return 1
     fi
-
-    # shellcheck disable=SC1090
-    source "$config_file"
+    # Store the current env so it can be re-applied after sourcing the config file
```
jharrod, line 327: Could we have just enforced either all params through the config file OR all params though args? I.e. just throw and error if someone supplies both? We are making an assumption here that if the user supplies both the one they actually want is in the args, which may or may not be true. We are guessing.

unified-installer/astra-unified-installer.sh
palicena: Trident would stay where it is, while the connector gets installed in the given NAMESPACE. Kind of weird behavior to be honest, but we don't currently "detect" to check if a connector is already installed.

unified-installer/astra-unified-installer.sh
```diff
@@ -188,6 +209,90 @@ get_configs() {
...
+  TRIDENT_ACP_IMAGE_TAG              The tag for the Trident ACP image.
+  CONNECTOR_OPERATOR_IMAGE_TAG       The tag for the Connector Operator image.
+
+  ----- Connector Configuration
+  *ASTRA_CONTROL_URL                 The Astra Control URL. Required, provided by the Astra Control UI.
+  *ASTRA_API_TOKEN                   The Astra API token. Required, provided by the Astra Control UI.
+  *ASTRA_ACCOUNT_ID                  The Astra account ID. Required, provided by the Astra Control UI.
+  *ASTRA_CLOUD_ID                    The Astra cloud ID. Required, provided by the Astra Control UI.
+  *ASTRA_CLUSTER_ID                  The Astra cluster ID. Required, provided by the Astra Control UI.
+  CONNECTOR_HOST_ALIAS_IP            Sets a host alias for the Astra Connector pod.
```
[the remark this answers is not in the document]
palicena: The host alias is a ["native" kubernetes pod field](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/) already (though I'm not sure that's what we actually use), so it may not be as cryptic as one might think. But yeah, not sure how else to word this.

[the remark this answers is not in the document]
palicena: Maybe 
```
Sets a host alias for the Astra Connector pod when connecting to the given ASTRA_CONTROL_URL.
```
?

unified-installer/astra-unified-installer.sh
palicena: It's not really a guess so much as common practice. Almost all CLIs behave that way as far as I know.

unified-installer/astra-unified-installer.sh
jharrod: would it install 2 connectors if the user entered a diff namespace than when initially installed?

unified-installer/astra-unified-installer.sh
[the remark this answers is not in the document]
jharrod: I like that, but maybe change "when" to "for"? 
"Sets a host alias for the Astra Connector pod for connecting to the given ASTRA_CONTROL_URL."

unified-installer/astra-unified-installer.sh
palicena: Yes, unfortunately. This is kind of a known issue that we deliberately planned on not fixing for this release since it's unlikely to happen I think, but it might be a good idea to add a note in the help description about it. I'll see what I can come up with.

unified-installer/astra-unified-installer.sh
jharrod: I don't think a customer would know or assume that. Most CLIs use --option=value type of args which aren't env based. I just thought this would clean the code up a bit and also ensures the customer knows what args they're actually using.

palicena: Sure, there's definitely the distinction between options/env vars to consider which might make things weird (since env vars can be inherited by the script without the user realizing -- though this can still happen even without this change), but I think it's still better than forcing one or the other.

We can't really force the config file unless we expand the UI instructions, and we can't force CLI args without removing crucial functionality that most customers will want.

The former (forcing the config file) is definitely viable, but unless you're willing to go to PMs about this and get them to reconsider, I don't think it's really negotiable at this stage in the process.

unified-installer/astra-unified-installer.sh
palicena: After looking into this, the changes were made on the astra-connector side, but we still need to tweak the script to support optional CLOUD_ID/CLUSTER_ID. I created a 1pt card to take care of this: https://jira.ngage.netapp.com/browse/ASTRACTL-33865

unified-installer/astra-unified-installer.sh
```diff
@@ -188,6 +210,93 @@ get_configs() {
...
+  CONFIG_FILE                        Path to a configuration file. Overrides environment variables specified via command line.
+  DRY_RUN                            Skips applying generated resources when set to true. Default is false.
+  SKIP_IMAGE_CHECK                   Skips checking if images exist when set to true. Default is false.
+  SKIP_ASTRA_CHECK                   Skips all checks requiring a connection to Astra Control when set to true. Default is false.
+  SKIP_TLS_VALIDATION                Skips TLS validation for all requests to Astra Control, including the Connector (unless CONNECTOR_SKIP_TLS_VALIDATION is set) when set to true. Default is false.
+  DISABLE_PROMPTS                    Skips all prompts, answering 'yes' by default when set to true. Default is false.
+  DO_NOT_MODIFY_EXISTING_TRIDENT     Prevents any and all modification to the existing Trident installation (if any). Required if DISABLE_PROMPTS is true, otherwise defaults to false.
+
+  ----- General Configuration
+  COMPONENTS                         Determines what will be installed/upgraded. Default is ALL_ASTRA_CONTROL.
```
josecrosa, line 242: Can we list what the options are?

jharrod (approved):

PR #352, opened 2024-06-04
ASTRACTL-32772: rework Trident upgrade logic, better error handling

palicena:
- Reworked and simplified Trident upgrade logic
- Added better error handling for all `kubectl get` commands; we'll now be able to differentiate connection/timeout errors VS not found, and users will get much better feedback when something does go wrong


unified-installer/astra-unified-installer.sh
```diff
@@ -1097,10 +1097,9 @@ get_registry_credentials_from_pull_secret() {
     # 6. Get the first matching entry's `.value.auth`, which would be `B64_ENCODED_USERNAME_PASSWORD` here
     # 7. The `B64_ENCODED_USERNAME_PASSWORD` can then be passed to curl via
     # `-H 'Authorization: Basic B64_ENCODED_USERNAME_PASSWORD'` without having to decode it first.
-    local -r contents="$(kubectl get secret "$pull_secret" -n "$namespace" -o json 2> /dev/null)"
+    local -r contents="$(k8s_get_resource "secret/$pull_secret" "$namespace" "json")"
```
palicena, line 1100: You'll see quite a few of these changes; most `kubectl get` commands now instead use `k8s_get_resource` or `k8s_resource_exists`, which actually handles kubectl errors instead of just sending them into the void.

unified-installer/astra-unified-installer.sh
```diff
@@ -1629,12 +1683,17 @@ step_check_k8s_version_in_range() {
 
     logheader $__DEBUG "Checking if Kubernetes version is within range ($minimum <=> $maximum)..."
 
-    local current; current="$(kubectl version -o json | jq -r '.serverVersion.gitVersion')"
+    local current; current="$(kubectl version -o json 2> "$__ERR_FILE" | jq -r '.serverVersion.gitVersion' 2> /dev/null)"
+    local -r captured_err="$(get_captured_err)"
+    if [ -z "$current" ] || [ "$current" == null ] || [ -n "$captured_err" ]; then
+        add_problem "Failed to get your cluster's Kubernetes version: $captured_err"
+        return 1
+    fi
```
palicena, line 1691: Since this is not a `kubectl get` and is only called once, the error is handled here instead of using `k8s_get_resource` or `k8s_resource_exists` (or making a new function).

unified-installer/astra-unified-installer.sh
```diff
@@ -1647,20 +1706,28 @@ step_check_k8s_permissions() {
...
+    if [ "$has_permission" == "yes" ]; then
         logdebug "k8s permissions: OK"
+        return 0
+    elif [ "$has_permission" != "yes" ]; then
+        add_problem "Kubernetes user does not have admin privileges"
+    elif [ -n "$captured_err" ]; then
+        add_problem "Failed to check if Kubernetes user has admin privilege: $captured_err"
     else
-        add_problem "k8s permissions: user does not have admin privileges" "Kubernetes user does not have admin privileges"
+        add_problem "Failed to check if Kubernetes user has admin privilege: unknown error"
```
palicena, line 1719: Similar to above, this `kubectl auth can-i` is pretty unique so we just handle the errors in-line instead of making a new function.

unified-installer/astra-unified-installer.sh
```diff
@@ -2350,6 +2416,7 @@ step_apply_resources() {
     # Apply operator resources
     logdebug "apply operator resources"
     local output=""
+    local captured_err=""
```
palicena, line 2419: This variable declaration was missing, so the call 3 lines down was putting the value in a global variable instead.

unified-installer/astra-unified-installer.sh
```diff
@@ -2552,43 +2619,28 @@ if trident_will_be_installed_or_modified; then
...
+                    trident_operator_image_needs_upgraded && step_generate_trident_operator_patch
+                elif prompt_user_yes_no "Would you like to upgrade Trident?"; then
+                    step_generate_torc_patch "${_generate_torc_args[@]}"
+                    trident_operator_image_needs_upgraded && step_generate_trident_operator_patch
+                else
+                    _msg="You have chosen to use a version of Trident that is not supported with the current version"
+                    _msg+=" of Astra Control. This may result in some App Data Management operations not functioning"
+                    _msg+=" correctly or being blocked within Astra Control. It is highly recommended to upgrade"
+                    _msg+=" Trident to ensure compatibility and proper functionality."
+                    logwarn "$_msg"
```
palicena, line 2643: Some of the Trident upgrade flow got condensed, while some of it was trimmed out since I realized it was being taken care of by code downstream (mainly the ACP-related deleted code below this comment).

palicena: 👍 I'll see about changing the messaging in the script as a whole, shouldn't be too terrible.

zachatnetapp (approved):

PR #358, opened 2024-06-05
ASTRACTL-33839

zachatnetapp:
This covers the following scenarios:

* add a k create secret command for trident-acp secret in the case where trident-acp needs to be enabled or upgraded
* if NAMESPACE is empty and components includes trident and connector, we now create a second secret for trident-acp

Both of these cases are assuming that the our ACS registry (cr.astra.netapp.io) is being used. If not, it is up to the script user to handle their own secrets.

+ Some other miscellaneous fixes that were already existing here https://github.com/NetApp/astra-connector-operator/pull/337 and that I needed for my changes to work right.

### DIFF
--- a/unified-installer/astra-unified-installer.sh
+++ b/unified-installer/astra-unified-installer.sh
@@ -1398,8 +1398,17 @@ step_check_config() {
         if [ -z "$NAMESPACE" ]; then
             prompt_user "NAMESPACE" "NAMESPACE is required when specifying an IMAGE_PULL_SECRET. Please enter the namespace:"
         fi
-        add_to_config_builder "IMAGE_PULL_SECRET"
+    elif config_has_at_least_one_custom_registry_or_repo; then
+        local custom_reg_warning="We detected one or more custom registry or repo values"
+        custom_reg_warning+=", but no IMAGE_PULL_SECRET was specified. If any of your images are hosted in a private"
+        custom_reg_warning+=" registry, an image pull secret will need to be created and IMAGE_PULL_SECRET set."
+        if prompts_disabled; then
+            logwarn "$custom_reg_warning"
+        elif prompt_user_yes_no "$custom_reg_warning${__NEWLINE}Would you like to specify a pull secret now?"; then
+            prompt_user "IMAGE_PULL_SECRET" "Enter a value for IMAGE_PULL_SECRET: "
+        fi
     fi
+    add_to_config_builder "IMAGE_PULL_SECRET"
     add_to_config_builder "NAMESPACE"
 
     if prompts_disabled; then
@@ -2238,7 +2247,7 @@ step_add_labels_to_kustomization() {
     logheader $__DEBUG "Adding labels to kustomization and crs file"
 
     local -r content="labels:${__NEWLINE}- pairs:${__NEWLINE}${processed_labels}"
-    insert_into_file_after_pattern "${kustomization_file}" "kind:" "${content}"
+    insert_into_file_after_pattern "${kustomization_file}" "kind: Kustomization" "${content}"
 
     logdebug "kustomization labels: OK"
 }
--- a/unified-installer/astra-unified-installer.sh
+++ b/unified-installer/astra-unified-installer.sh
@@ -1786,6 +1786,87 @@ EOF
     fi
 }
 
+step_kustomize_global_pull_secret_if_needed() {
+    local -r global_pull_secret="${1:-""}"
+    local -r kustomization_file="${2}"
+    local -r kustomization_dir="$(dirname "$kustomization_file")"
+    local -r connector_namespace="$(get_connector_namespace)"
+    local -r trident_namespace="$(get_trident_namespace)"
+    local -r connector_registry="$(join_rpath "$CONNECTOR_IMAGE_REGISTRY" "$(get_base_repo "$CONNECTOR_IMAGE_REPO")")"
+    local -r trident_acp_registry="$(join_rpath "$TRIDENT_ACP_IMAGE_REGISTRY" "$(get_base_repo "$TRIDENT_ACP_IMAGE_REPO")")"
+    local -r encoded_creds=$(echo -n "$ASTRA_ACCOUNT_ID:$ASTRA_API_TOKEN" | base64)
+
+    [ -z "$kustomization_file" ] && fatal "no kustomization file given"
+    [ ! -f "$kustomization_file" ] && fatal "kustomization file '$kustomization_file' does not exist"
+
+       # SECRET GENERATOR
+    cat <<EOF >> "$kustomization_file"
+generatorOptions:
+  disableNameSuffixHash: true
+secretGenerator:
+- name: astra-api-token
+  namespace: "${connector_namespace}"
+  literals:
+  - apiToken="${ASTRA_API_TOKEN}"
+EOF
+    if [ -z "$global_pull_secret" ]; then
+        # if image pull secret is empty, set same name for connector and trident secret so torc patch works as expected
+        IMAGE_PULL_SECRET="astra-regcred"
+        if components_include_connector; then
+            cat <<EOF >> "$kustomization_file"
+- name: "${IMAGE_PULL_SECRET}"
+  namespace: "${connector_namespace}"
+  type: kubernetes.io/dockerconfigjson
+  literals:
+  - |
+    .dockerconfigjson={
+      "auths": {
+        "${connector_registry}": {
+          "username": "$ASTRA_ACCOUNT_ID",
+          "password": "$ASTRA_API_TOKEN",
+          "auth": "${encoded_creds}"
+        }
+      }
+    }
+EOF
+            logdebug "$kustomization_file: added connector secret to namespace $connector_namespace"
+        fi
+    
+        if components_include_trident && [ "$trident_namespace" != "$connector_namespace" ]; then
+            cat <<EOF >> "$kustomization_file"
+- name: "${IMAGE_PULL_SECRET}"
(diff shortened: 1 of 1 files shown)


unified-installer/astra-unified-installer.sh
```diff
@@ -2436,13 +2540,17 @@ if trident_will_be_installed_or_modified; then
             # Enable ACP if needed (includes ACP upgrade)
             if ! acp_is_enabled; then
                 if config_acp_image_is_custom || prompt_user_yes_no "Would you like to enable ACP?"; then
+                    # create trident-acp secret
+                    kubectl create secret docker-registry "$IMAGE_PULL_SECRET" --docker-username="$ASTRA_ACCOUNT_ID" --docker-password="$ASTRA_API_TOKEN" -n trident --docker-server="$TRIDENT_ACP_IMAGE_REGISTRY"
```
palicena, line 2544: Instead of creating the resource immediately, we should instead generate a file with the secret YAML in it which then gets applied in `step_apply_resources`. This would only require adding the `--dry-run=client` and `-o yaml` options:
```
>> kubectl create secret docker-registry test-pull-secret -n astra --docker-username='<account_id>' --docker-password='<api_token>' --docker-server='<acs_registry>' --dry-run=client -o yaml

apiVersion: v1
data:
  .dockerconfigjson: eyJhdXRocyI6eyJcdTAwM2NhY3NfcmVnaXN0cnlcdTAwM2UiOnsidXNlcm5hbWUiOiJcdTAwM2NhY2NvdW50X2lkXHUwMDNlIiwicGFzc3dvcmQiOiJcdTAwM2NhcGlfdG9rZW5cdTAwM2UiLCJhdXRoIjoiUEdGalkyOTFiblJmYVdRK09qeGhjR2xmZEc5clpXNCsifX19
kind: Secret
metadata:
  creationTimestamp: null
  name: test-pull-secret
  namespace: astra
type: kubernetes.io/dockerconfigjson
```
You can then save this into a file in the operators folder. Tracking the file name will likely require adding a new 
`__GENERATED` variable alongside the others:
```
readonly __GENERATED_CRS_DIR="./astra-generated"
readonly __GENERATED_CRS_FILE="$__GENERATED_CRS_DIR/crs.yaml"
readonly __GENERATED_OPERATORS_DIR="$__GENERATED_CRS_DIR/operators"
readonly __GENERATED_KUSTOMIZATION_FILE="$__GENERATED_OPERATORS_DIR/kustomization.yaml"
readonly __GENERATED_PATCHES_TORC_FILE="$__GENERATED_CRS_DIR/post-deploy-patches_torc"
readonly __GENERATED_PATCHES_TRIDENT_OPERATOR_FILE="$__GENERATED_OPERATORS_DIR/post-deploy-patches_trident-operator"
# add  __GENERATED_TRIDENT_ACP_SECRET_FILE or something like that
```

unified-installer/astra-unified-installer.sh
```diff
@@ -2478,11 +2528,10 @@ step_generate_trident_operator_patch() {
         if echo "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" | grep -q "^${IMAGE_PULL_SECRET}$" &> /dev/null; then
             logdebug "image pull secret '$IMAGE_PULL_SECRET' already present in trident-operator"
         else
-            local -r secret_obj='{"name": "'"$IMAGE_PULL_SECRET"'"}'
             if [ -z "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" ]; then
-                patch_list+=',{"op":"replace","path":"/spec/template/spec/imagePullSecrets","value":['"$secret_obj"']}'
```
zachatnetapp, line 2483: This is just for consistency. We don't have this $secret_obj variable in step_generate_torc_patch(), so I'd rather we do the same thing in both places. This came up with I was solving a merge conflict, otherwise I would not have even thought to touch this.

unified-installer/astra-unified-installer.sh
zachatnetapp: Done

unified-installer/astra-unified-installer.sh
```diff
@@ -2531,7 +2559,7 @@ step_generate_torc_patch() {
             if [ -z "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" ]; then
                 torc_patch_list+='{"op":"replace","path":"/spec/imagePullSecrets","value":['"$IMAGE_PULL_SECRET"']},'
             else
-                torc_patch_list+='{"op":"add","path":"/spec/imagePullSecrets/-","value":"'"$IMAGE_PULL_SECRET"'"},'
```
palicena, line 2534: Those double quotes are needed since the value is a JSON string. It's a little convoluted, but the jist of it is:
1. The first double quotes `"` are taken literally because they are within a set of single quotes `'`
2. The single quote `'` after this is closing the set of single quotes that was started at `torc_patch_list+='`
3. The second double quotes `"` are bash quotes to prevent globbing/word splitting (see https://github.com/koalaman/shellcheck/wiki/SC2086) from when `$IMAGE_PULL_SECRET` gets expanded.

unified-installer/astra-unified-installer.sh
```diff
@@ -2841,14 +2874,18 @@ if trident_will_be_installed_or_modified; then
         if components_include_acp; then
             # Enable ACP if needed (includes ACP upgrade)
             if ! acp_is_enabled; then
-                if config_acp_image_is_custom || prompt_user_yes_no "Would you like to enable Astra Control Provisioner?"; then
+                if config_acp_image_is_custom || prompt_user_yes_no "Would you like to enable ACP?"; then
```
palicena, line 2877: `ACP` -> `Astra Control Provisioner`

palicena, line 2877:
```suggestion
                if config_acp_image_is_custom || prompt_user_yes_no "Would you like to enable Astra Control Provisioner?"; then
```

unified-installer/astra-unified-installer.sh
```diff
@@ -2841,14 +2874,18 @@ if trident_will_be_installed_or_modified; then
...
+                    # create trident-acp secret
+                    kubectl create secret docker-registry "$IMAGE_PULL_SECRET" --docker-username="$ASTRA_ACCOUNT_ID" --docker-password="$ASTRA_API_TOKEN" -n trident --docker-server="$TRIDENT_ACP_IMAGE_REGISTRY" --dry-run=client -o yaml > "$__GENERATED_TRIDENT_ACP_SECRET_FILE"
                     step_generate_torc_patch "$_EXISTING_TORC_NAME" "" "$(get_config_acp_image)" "true"
                 else
                     loginfo "Astra Control Provisioner will not be enabled."
                 fi
             # ACP upgrade (ACP already enabled)
             elif acp_image_needs_upgraded; then
-                if config_acp_image_is_custom || prompt_user_yes_no "Would you like to upgrade Astra Control Provisioner?"; then
+                if config_acp_image_is_custom || prompt_user_yes_no "Would you like to upgrade ACP?"; then
```
palicena, line 2886:
```suggestion
                if config_acp_image_is_custom || prompt_user_yes_no "Would you like to upgrade Astra Control Provisioner?"; then
```

unified-installer/astra-unified-installer.sh
zachatnetapp: Ohhhhh, got it. Ill fix that.

zachatnetapp: Fixed, good catch. Thanks

unified-installer/astra-unified-installer.sh
```diff
@@ -2841,14 +2874,18 @@ if trident_will_be_installed_or_modified; then
         if components_include_acp; then
             # Enable ACP if needed (includes ACP upgrade)
             if ! acp_is_enabled; then
-                if config_acp_image_is_custom || prompt_user_yes_no "Would you like to enable Astra Control Provisioner?"; then
+                if config_acp_image_is_custom || prompt_user_yes_no "Would you like to enable ACP?"; then
+                    # create trident-acp secret
+                    kubectl create secret docker-registry "$IMAGE_PULL_SECRET" --docker-username="$ASTRA_ACCOUNT_ID" --docker-password="$ASTRA_API_TOKEN" -n trident --docker-server="$TRIDENT_ACP_IMAGE_REGISTRY" --dry-run=client -o yaml > "$__GENERATED_TRIDENT_ACP_SECRET_FILE"
```
palicena, line 2879: Since this is duplicated as is, might want to just put it in a function. You can then also add error handling to it, in the event creating the secret fails. For example, error handling might look something like this:
```
    local -r secret_yaml="$(kubectl create secret docker-registry "$IMAGE_PULL_SECRET" --docker-username="$ASTRA_ACCOUNT_ID" --docker-password="$ASTRA_API_TOKEN" -n trident --docker-server="$TRIDENT_ACP_IMAGE_REGISTRY" --dry-run=client -o yaml 2> "$__ERR_FILE"
    local -r captured_err="$(get_captured_err)"
    if [ -z "$secret_yaml" ] || [ -n "$captured_err" ]; then
        add_problem "Failed to generate secret for ACS registry '$TRIDENT_ACP_IMAGE_REGISTRY': $captured_err"
        return 1
    fi
```
- The ` 2> "$__ERR_FILE"` redirects the command's error messages (which goes through stderr, which is indicated by the `2`, while `>` is the redirection) into an error file we maintain in the script for exactly this purpose. 
- The next line, `local -r captured_err="$(get_captured_err)"`, simply gets the value we saved in that error file.
- We then check if the secret was actually generated or if we captured an error

palicena (approved):
👌

PR #362, opened 2024-06-05
ASTRACTL-32772: bug fixes

palicena:
Two fixes for random bugs I found while testing some unrelated changes.


unified-installer/astra-unified-installer.sh
```diff
@@ -1707,7 +1707,12 @@ step_check_config() {
     # Env vars with special conditions
     if [ -n "$IMAGE_PULL_SECRET" ]; then
         if [ -z "$NAMESPACE" ]; then
-            prompt_user "NAMESPACE" "NAMESPACE is required when specifying an IMAGE_PULL_SECRET. Please enter the namespace:"
+            local -r ns_warning="NAMESPACE is required when specifying an IMAGE_PULL_SECRET"
+            if prompts_disabled; then
+                add_problem "$ns_warning"
+            else
+                prompt_user "NAMESPACE" "$ns_warning. Please enter the namespace:"
+            fi
```
palicena, line 1715: With prompts disabled, we were letting the user proceed without a NAMESPACE which would cause a fatal error down the road since we were trying to do operations involving an IMAGE_PULL_SECRET and relied on NAMESPACE also being set.

unified-installer/astra-unified-installer.sh
```diff
@@ -2656,12 +2661,17 @@ step_apply_resources() {
...
+        if [ -n "$captured_err" ]; then
+            while IFS= read -r line; do
+                if echo "$line" | grep -q "Warning:"; then
+                    logdebug "captured warning when applying kustomize resources:${__NEWLINE}$captured_err"
+                elif echo "$line" | grep -q "no objects passed to apply"; then
+                    logdebug "no kustomize resources to apply, skipping"
+                elif [ -z "$output" ] || [ -n "$line" ]; then
+                    add_problem "Failed to apply kustomize resources: $line"
+                fi
+            done < <(echo "$captured_err")
```
palicena, line 2674: If the `apply` command has both a set of warnings and errors, the warning would basically "hide" the actual errors since we were checking the whole string at once instead of line by line. We'd then continue with the rest of the script even though the apply failed.

unified-installer/astra-unified-installer.sh
zachatnetapp: Nice catch!

unified-installer/astra-unified-installer.sh
zachatnetapp: Another good catch

PR #363, opened 2024-06-06
ASTACTL-32772: kubectl version/permission fixes

palicena:
- Fix `kubectl get version` warnings being interpreted as errors for `
- Fix `kubectl auth can-i` errors being interpreted as the user not having admin access

### DIFF
--- a/unified-installer/astra-unified-installer.sh
+++ b/unified-installer/astra-unified-installer.sh
@@ -725,7 +725,8 @@ loginfo() {
 }
 
 logwarn() {
-    log_at_level $__WARN "WARNING: $1"
+    local -r warning="${1#WARNING: }"
+    log_at_level $__WARN "WARNING: $warning"
 }
 
 logerror() {
@@ -1767,10 +1768,14 @@ step_check_k8s_version_in_range() {
     logheader $__DEBUG "Checking if Kubernetes version is within range ($minimum <=> $maximum)..."
 
     local current; current="$(kubectl version -o json 2> "$__ERR_FILE" | jq -r '.serverVersion.gitVersion' 2> /dev/null)"
-    local -r captured_err="$(get_captured_err)"
-    if [ -z "$current" ] || [ "$current" == null ] || [ -n "$captured_err" ]; then
+    local captured_err; captured_err="$(get_captured_err)"
+    if [ -z "$current" ] || [ "$current" == null ]; then
+        [ -z "$captured_err" ] && captured_err="unknown error"
         add_problem "Failed to get your cluster's Kubernetes version: $captured_err"
         return 1
+    fi
+    if echo "$captured_err" | starts_with "WARNING"; then
+        logwarn "$captured_err"
     fi
 
     current=${current#v}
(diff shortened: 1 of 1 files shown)


unified-installer/astra-unified-installer.sh
```diff
@@ -726,7 +726,8 @@ loginfo() {
 }
 
 logwarn() {
-    log_at_level $__WARN "WARNING: $1"
+    local -r warning="${1#WARNING: }"
+    log_at_level $__WARN "WARNING: $warning"
```
palicena, line 730: This is just to prevent having some kubectl warnings say `WARNING: WARNING: blablabla`.

unified-installer/astra-unified-installer.sh
```diff
@@ -1802,11 +1803,15 @@ step_check_k8s_version_in_range() {
     logheader $__DEBUG "Checking if Kubernetes version is within range ($minimum <=> $maximum)..."
 
     local current; current="$(kubectl version -o json 2> "$__ERR_FILE" | jq -r '.serverVersion.gitVersion' 2> /dev/null)"
-    local -r captured_err="$(get_captured_err)"
-    if [ -z "$current" ] || [ "$current" == null ] || [ -n "$captured_err" ]; then
+    local captured_err; captured_err="$(get_captured_err)"
+    if [ -z "$current" ] || [ "$current" == null ]; then
+        [ -z "$captured_err" ] && captured_err="unknown error"
```
palicena, line 1808: Previously, the check for `captured_err` was part of the first `if`, but warnings actually go through stderr (same as errors), so the following was being interpreted as an error:
```
WARNING: We cannot verify the version of the custom Astra Trident image you provided (docker.repo.eng.netapp.com/palice/trident:dev. Astra Control Provisioner support (23.10+) is not guaranteed.
```
This caused a successful check to instead manifest as a failure.

unified-installer/astra-unified-installer.sh
```diff
@@ -1802,11 +1803,15 @@ step_check_k8s_version_in_range() {
...
         add_problem "Failed to get your cluster's Kubernetes version: $captured_err"
         return 1
     fi
+    if echo "$captured_err" | starts_with "WARNING"; then
+        logwarn "$captured_err"
+    fi
```
palicena, line 1814: Bubble up the warning so the user is aware of it:
```
--- Checking if all required tools are installed...

--- Checking your Kubernetes version...
WARNING: version difference between client (1.29) and server (1.24) exceeds the supported minor version skew of +/-1

--- Checking for volumesnapshotclasses crd...
```

unified-installer/astra-unified-installer.sh
```diff
@@ -1829,7 +1834,7 @@ step_check_k8s_permissions() {
     if [ "$has_permission" == "yes" ]; then
         logdebug "k8s permissions: OK"
         return 0
-    elif [ "$has_permission" != "yes" ]; then
+    elif echo "$has_permission" | grep -q "no"; then
```
palicena, line 1837: This check was too loose; an empty value would manifest as the user not having permission, when it actually means there's an error.

jharrod (approved):

PR #364, opened 2024-06-10
ASTRACTL-33567: split NAMESPACE -> CONNECTOR/TRIDENT_NAMESPACE

palicena:
This PR adds `CONNECTOR_NAMESPACE` and `TRIDENT_NAMESPACE`. The original NAMESPACE variable is now just a high level "default", similar to IMAGE_REGISTRY, IMAGE_BASE_REPO, and so on.

There are two "major" changes that were done to achieve this:
- `operators/kustomization.yaml` is now split into two:
    - `operators/connector/kustomization.yaml` - contains all kustomizations related to the connector
    - `operators/trident/kustomization.yaml` - contains all kustomizations related to trident
- The **astra-regcred**, which is the default secret used for authenticating to the Astra registry and is generated by us, is now completely separate from **IMAGE_PULL_SECRET**. This means that:
    -  If we detect any of the image registries are of the Astra flavor, we create **astra-regcred** regardless of what's in **IMAGE_PULL_SECRET**
    - All resources will have their imagePullSecret references set appropriately (will both include **IMAGE_PULL_SECRET** if there is one as well as **astra-regcred** if we found an image hosted on an Astra registry)
    
Also contains a few bug fixes that I encountered while testing -- which, on that topic, included:
- Generic Greenfield test (connector + trident/acp)
- Generic Brownfield test (connector + trident upgrade)
- Generic Brownfield test with DO_NOT_MODIFY_EXISTING_TRIDENT (connector + existing trident left intact)
- Multiple specialized greenfield/brownfield scenarios to make sure pull secrets are being created, including manipulating the list of pull secrets of an existing installation to make sure we correctly add the IMAGE_PULL_SECRET or astra-regcred to the list (if they're mising) during an upgrade

### DIFF
--- a/unified-installer/astra-unified-installer.sh
+++ b/unified-installer/astra-unified-installer.sh
@@ -12,6 +12,9 @@
 _PROBLEMS=() # Any failed checks will be added to this array and the program will exit at specific checkpoints if not empty
 _CONFIG_BUILDER=() # Contains the fully resolved config to be output at the end during a dry run
 _KUBERNETES_VERSION=""
+# _RESOURCES_OR_PATCHES_HAVE_BEEN_APPLIED is to determine whether the script did anything or not, so we can
+# let the user if it was a no-op.
+_RESOURCES_OR_PATCHES_HAVE_BEEN_APPLIED="false"
 
 # _TRIDENT_COLLECTION_STEP_CALLED is used as a guardrail to prevent certain functions from being called before
 # existing Trident information (if any) has been collected.
@@ -65,6 +68,9 @@ readonly __COMPONENTS_ACP_ONLY="ACP_ONLY"
 readonly __COMPONENTS_VALID_VALUES=("$__COMPONENTS_ALL_ASTRA_CONTROL" "$__COMPONENTS_TRIDENT_AND_ACP" \
     "$__COMPONENTS_TRIDENT_ONLY" "$__COMPONENTS_ACP_ONLY")
 
+readonly __ASTRA_API_TOKEN_PULL_SECRET="astra-api-token"
+readonly __ASTRA_REGISTRY_IMAGE_PULL_SECRET="astra-regcred"
+
 readonly __DEFAULT_DOCKER_HUB_IMAGE_REGISTRY="docker.io"
 readonly __DEFAULT_DOCKER_HUB_IMAGE_BASE_REPO="netapp"
 readonly __DEFAULT_ASTRA_IMAGE_REGISTRY="cr.astra.netapp.io"
@@ -87,10 +93,15 @@ readonly __PRODUCTION_AUTOSUPPORT_URL="https://support.netapp.com/put/AsupPut"
 readonly __GENERATED_CRS_DIR="./astra-generated"
 readonly __GENERATED_CRS_FILE="$__GENERATED_CRS_DIR/crs.yaml"
 readonly __GENERATED_OPERATORS_DIR="$__GENERATED_CRS_DIR/operators"
+readonly __GENERATED_OPERATORS_TRIDENT_DIR="$__GENERATED_CRS_DIR/operators/trident"
+readonly __GENERATED_OPERATORS_CONNECTOR_DIR="$__GENERATED_CRS_DIR/operators/connector"
 readonly __GENERATED_KUSTOMIZATION_FILE="$__GENERATED_OPERATORS_DIR/kustomization.yaml"
+readonly __GENERATED_TRIDENT_KUSTOMIZATION_FILE="$__GENERATED_OPERATORS_TRIDENT_DIR/kustomization.yaml"
+readonly __GENERATED_CONNECTOR_KUSTOMIZATION_FILE="$__GENERATED_OPERATORS_CONNECTOR_DIR/kustomization.yaml"
+
 readonly __GENERATED_PATCHES_TORC_FILE="$__GENERATED_CRS_DIR/post-deploy-patches_torc"
 readonly __GENERATED_PATCHES_TRIDENT_OPERATOR_FILE="$__GENERATED_OPERATORS_DIR/post-deploy-patches_trident-operator"
-readonly __GENERATED_TRIDENT_ACP_SECRET_FILE="$__GENERATED_OPERATORS_DIR/trident-acp-secret.yaml"
+readonly __GENERATED_TRIDENT_ACP_SECRET_FILE="$__GENERATED_CRS_DIR/trident-acp-secret.yaml"
 
 readonly __DEBUG=10
 readonly __INFO=20
@@ -146,6 +157,8 @@ get_configs() {
     COMPONENTS="${COMPONENTS:-$__COMPONENTS_ALL_ASTRA_CONTROL}" # Determines what we'll install/upgrade
     IMAGE_PULL_SECRET="${IMAGE_PULL_SECRET:-}" # TODO ASTRACTL-32772: skip prompt if IMAGE_REGISTRY is default
     NAMESPACE="${NAMESPACE:-}" # Overrides EVERY resource's namespace (for fresh installs only, not upgrades)
+        CONNECTOR_NAMESPACE="$(get_connector_namespace)"
+        TRIDENT_NAMESPACE="$(get_trident_namespace)"
     LABELS="${LABELS:-}"
     # SKIP_TLS_VALIDATION will skip TLS validation for all requests made during the script.
     SKIP_TLS_VALIDATION="${SKIP_TLS_VALIDATION:-"false"}"
@@ -249,7 +262,9 @@ Optional Environment Variables:
   ----- General Configuration
   COMPONENTS                         One of [${__COMPONENTS_VALID_VALUES[*]}]. Determines what will be installed/upgraded. Default is ALL_ASTRA_CONTROL.
   IMAGE_PULL_SECRET                  Image pull secret for the Docker registry.
-  NAMESPACE                          Overrides EVERY resource's namespace (for fresh installs only, not upgrades).
+  NAMESPACE                          Sets the namespace for all applied resources. If an existing Trident installation is found, its namespace will instead be used for all Trident operations. Overridden by _NAMESPACE values below.
+  CONNECTOR_NAMESPACE                Sets the namespace for all applied Astra Connector resources. Defaults to 'astra-connector'.
+  TRIDENT_NAMESPACE                  Sets the namespace for all applied Astra Trident resources. If an existing Trident installation is found, its namespace will instead be used for all Trident operations. Defaults to 'trident'.
   LABELS                             Labels to be added to the generated resources (disclaimer: does not apply labels to resources created by the operators).
   RESOURCE_LIMITS_PRESET             Resource limit preset to use. Overridden by RESOURCE_LIMITS_CUSTOM_CPU and RESOURCE_LIMITS_CUSTOM_MEMORY (disclaimer: resource limits not currently supported by Astra Trident).
   RESOURCE_LIMITS_CUSTOM_CPU         Custom CPU resource limit. Plain number.
@@ -328,6 +343,7 @@ ingest_config_string() {
 load_config_from_file_if_given() {
     local config_file=$1
     local api_token=$ASTRA_API_TOKEN
+    local ingested_api_token_through_config_file="false"
     local token_warning="We detected that your ASTRA_API_TOKEN was provided through the CONFIG_FILE,"
     token_warning+=" which may pose a security risk! Make sure to store the configuration file in a secure location,"
     token_warning+=" or consider moving the API token out of the file and providing it through the command line only when needed."
@@ -357,17 +373,20 @@ load_config_from_file_if_given() {
 
     # Set config file values first
     ingest_config_string "$(cat "$config_file")"
-    set_log_level
 
     # Check if api token was populated after sourcing config file
     if [ "$api_token" != "$ASTRA_API_TOKEN" ]; then
-        logwarn "$token_warning"
+        ingested_api_token_through_config_file="true"
     fi
 
     # Set previous env second to allow vars provided through the command line to take priority
     ingest_config_string "$previous_env_double_quoted"
     set_log_level
 
+    # Print the token warning AFTER applying the previous env to make sure we respect the final resolved log level
+    if [ "$ingested_api_token_through_config_file" == "true" ]; then
+        logwarn "$token_warning"
+    fi
     logheader $__DEBUG "Loaded configuration from file: $config_file"
 }
 
@@ -413,15 +432,15 @@ components_include_acp() {
 }
 
 get_trident_namespace() {
-    echo "${_EXISTING_TRIDENT_NAMESPACE:-"${NAMESPACE:-"${__DEFAULT_TRIDENT_NAMESPACE}"}"}"
+    echo "${_EXISTING_TRIDENT_NAMESPACE:-"${TRIDENT_NAMESPACE:-"${NAMESPACE:-"$__DEFAULT_TRIDENT_NAMESPACE"}"}"}"
 }
 
 get_connector_operator_namespace() {
-    echo "${NAMESPACE:-"${__DEFAULT_CONNECTOR_OPERATOR_NAMESPACE}"}"
+    get_connector_namespace
 }
 
 get_connector_namespace() {
-    echo "${NAMESPACE:-"${__DEFAULT_CONNECTOR_NAMESPACE}"}"
+    echo "${CONNECTOR_NAMESPACE:-"${NAMESPACE:-"$__DEFAULT_CONNECTOR_NAMESPACE"}"}"
 }
 
 existing_trident_can_be_modified() {
@@ -753,6 +772,19 @@ fatal() {
     exit 1
 }
 
+# remove_kubectl_warnings_from_error_output will output kubectl warnings to stderr and actual errors
+# to stdout, allowing one to examine proper errors separately from warnings.
+# Usage: local -r actual_errors="$(echo "$my_error_output" | remove_kubectl_warnings_from_error_output)"
+remove_kubectl_warnings_from_error_output() {
+    while IFS= read -r line; do
+        if echo "$line" | grep -q "Warning:"; then
+            logdebug "captured kubectl apply warning:${__NEWLINE}$line" 1>&2
+        else
+            echo "$line"
+        fi
+    done
+}
+
 insert_into_file_after_pattern() {
     local -r filepath="$1"
     local -r pattern="$2"
@@ -790,6 +822,12 @@ append_lines_to_file() {
 
     [ -z "$file" ] && fatal "no file given"
     [ "${#lines[@]}" -eq 0 ] && fatal "no lines given"
+
+    if [ ! -f "$file" ]; then
+        if ! touch "$file" 1> /dev/null; then
+            fatal "failed to create file '$file'"
+        fi
+    fi
 
     for line in "${lines[@]}"; do
         echo "$line"
@@ -905,6 +943,30 @@ join_rpath() {
     echo "$joined"
 }
 
+join_str() {
+    local -r separator="${1}"
+    local -ar strings_to_join=("${@:2}")
+    [ "${#strings_to_join[@]}" -eq 0 ] && return 0
+    [ "${#strings_to_join[@]}" -eq 1 ] && echo "$2" && return 0
+    [ -z "$separator" ] && fatal "no separator given"
+
+    local joined=""
+    for str in "${strings_to_join[@]}" ; do
+        [ -n "$joined" ] && joined+="$separator"
+        joined+="$str"
+    done
+
+    echo "$joined"
+}
+
+# is_empty returns 0 (true) if no args or a single empty arg is provided.
+is_empty() {
+    local -r args=("${@}")
+    [ "${#args[@]}" -eq 0 ] && return 0
+    [ "${#args[@]}" -eq 1 ] && [ -z "${args[0]}" ] && return 0
+    return 1
+}
+
 # starts_with checks if a string starts with the given substring.
 # Usage: `echo "abc" | starts_with "ab"` returns 0 (true).
 starts_with() {
@@ -973,7 +1035,7 @@ get_base_repo() {
     local -r image_repo="$1"
     if [ -z "$image_repo" ]; then fatal "no image_repo given"; fi
 
-    local -r base="$(dirname "$image_repo")"
+    local base; base="$(dirname "$image_repo")"
     if [ "$base" == "." ]; then base="$image_repo"; fi
 
     echo "$base"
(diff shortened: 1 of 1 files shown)


unified-installer/astra-unified-installer.sh
```diff
@@ -12,6 +12,9 @@
 _PROBLEMS=() # Any failed checks will be added to this array and the program will exit at specific checkpoints if not empty
 _CONFIG_BUILDER=() # Contains the fully resolved config to be output at the end during a dry run
 _KUBERNETES_VERSION=""
+# _RESOURCES_OR_PATCHES_HAVE_BEEN_APPLIED is to determine whether the script did anything or not, so we can
+# let the user if it was a no-op.
+_RESOURCES_OR_PATCHES_HAVE_BEEN_APPLIED="false"
```
palicena, line 17: Messaging wasn't very clear when the script ended up doing nothing. This will be used to output a message at the very end telling the user if the script was a noop.

unified-installer/astra-unified-installer.sh
```diff
@@ -65,6 +68,9 @@ readonly __COMPONENTS_ACP_ONLY="ACP_ONLY"
 readonly __COMPONENTS_VALID_VALUES=("$__COMPONENTS_ALL_ASTRA_CONTROL" "$__COMPONENTS_TRIDENT_AND_ACP" \
     "$__COMPONENTS_TRIDENT_ONLY" "$__COMPONENTS_ACP_ONLY")
 
+readonly __ASTRA_API_TOKEN_PULL_SECRET="astra-api-token"
+readonly __ASTRA_REGISTRY_IMAGE_PULL_SECRET="astra-regcred"
```
palicena, line 72: Just making the names official since it was hardcoded in a few different places.

unified-installer/astra-unified-installer.sh
```diff
@@ -87,10 +93,15 @@ readonly __PRODUCTION_AUTOSUPPORT_URL="https://support.netapp.com/put/AsupPut"
...
+readonly __GENERATED_OPERATORS_TRIDENT_DIR="$__GENERATED_CRS_DIR/operators/trident"
+readonly __GENERATED_OPERATORS_CONNECTOR_DIR="$__GENERATED_CRS_DIR/operators/connector"
 readonly __GENERATED_KUSTOMIZATION_FILE="$__GENERATED_OPERATORS_DIR/kustomization.yaml"
+readonly __GENERATED_TRIDENT_KUSTOMIZATION_FILE="$__GENERATED_OPERATORS_TRIDENT_DIR/kustomization.yaml"
+readonly __GENERATED_CONNECTOR_KUSTOMIZATION_FILE="$__GENERATED_OPERATORS_CONNECTOR_DIR/kustomization.yaml"
+
 readonly __GENERATED_PATCHES_TORC_FILE="$__GENERATED_CRS_DIR/post-deploy-patches_torc"
 readonly __GENERATED_PATCHES_TRIDENT_OPERATOR_FILE="$__GENERATED_OPERATORS_DIR/post-deploy-patches_trident-operator"
-readonly __GENERATED_TRIDENT_ACP_SECRET_FILE="$__GENERATED_OPERATORS_DIR/trident-acp-secret.yaml"
+readonly __GENERATED_TRIDENT_ACP_SECRET_FILE="$__GENERATED_CRS_DIR/trident-acp-secret.yaml"
```
palicena, line 104: The main additions here are the specific `TRIDENT` and `CONNECTOR` directories and kustomization files. The split was necessary because namespace transformers don't have the ability to differentiate connector/trident resources. This means we have:

- `operators/kustomization.yaml` - includes the two next files as resources
    - `operators/connector/kustomization.yaml` - pulls in the astra-connector-operator kustomization file up in github
    - `operators/trident/kustomization.yaml` - pulls in the trident kustomization file up in github

unified-installer/astra-unified-installer.sh
```diff
@@ -357,17 +373,20 @@ load_config_from_file_if_given() {
 
     # Set config file values first
     ingest_config_string "$(cat "$config_file")"
-    set_log_level
 
     # Check if api token was populated after sourcing config file
     if [ "$api_token" != "$ASTRA_API_TOKEN" ]; then
-        logwarn "$token_warning"
```
palicena, line 364: This log was being called after we ingested the config file, but before we apply values given through the command line. This meant that, if the config file had `LOG_LEVEL=debug` but the command line had `LOG_LEVEL=info`, you would see this warning with the `|WARN |` prefix even though it shouldn't show with log level **info**.

unified-installer/astra-unified-installer.sh
```diff
@@ -413,15 +432,15 @@ components_include_acp() {
 }
 
 get_trident_namespace() {
-    echo "${_EXISTING_TRIDENT_NAMESPACE:-"${NAMESPACE:-"${__DEFAULT_TRIDENT_NAMESPACE}"}"}"
+    echo "${_EXISTING_TRIDENT_NAMESPACE:-"${TRIDENT_NAMESPACE:-"${NAMESPACE:-"$__DEFAULT_TRIDENT_NAMESPACE"}"}"}"
 }
 
 get_connector_operator_namespace() {
-    echo "${NAMESPACE:-"${__DEFAULT_CONNECTOR_OPERATOR_NAMESPACE}"}"
+    get_connector_namespace
```
palicena, line 439: `__DEFAULT_CONNECTOR_OPERATOR_NAMESPACE` didn't actually exist, and, as far as the script is concerned, this needs to always be in the connector namespace.

unified-installer/astra-unified-installer.sh
```diff
@@ -753,6 +772,19 @@ fatal() {
     exit 1
 }
 
+# remove_kubectl_warnings_from_error_output will output kubectl warnings to stderr and actual errors
+# to stdout, allowing one to examine proper errors separately from warnings.
+# Usage: local -r actual_errors="$(echo "$my_error_output" | remove_kubectl_warnings_from_error_output)"
+remove_kubectl_warnings_from_error_output() {
```
palicena, line 778: This is to fix a bug where some `kubectl apply` commands were reported as failed because kubectl warnings are output on the same channel as errors, making warnings appear as errors.

unified-installer/astra-unified-installer.sh
```diff
@@ -973,7 +1035,7 @@ get_base_repo() {
     local -r image_repo="$1"
     if [ -z "$image_repo" ]; then fatal "no image_repo given"; fi
 
-    local -r base="$(dirname "$image_repo")"
+    local base; base="$(dirname "$image_repo")"
```
palicena, line 1038: Small bug where `base` was set as readonly despite being modified just below.

unified-installer/astra-unified-installer.sh
```diff
@@ -1117,12 +1179,177 @@ is_astra_registry() {
     local -r registry_url="$1"
     if [ -z "$registry_url" ]; then fatal "no registry_url given"; fi
 
-    if echo "$registry" | grep -q "cr." && echo "$registry" | grep -q "astra"; then
+    if echo "$registry_url" | grep -q "cr." && echo "$registry_url" | grep -q "astra"; then
```
palicena, line 1182: Bug; variable got renamed at some point but these references weren't changed alongside it.

unified-installer/astra-unified-installer.sh
```diff
@@ -1117,12 +1179,177 @@ is_astra_registry() {
...
         return 0
     fi
     return 1
 }
 
+# find_astra_registry loops through the given list of registries and returns the first one which matches the
+# pattern of an Astra registry.
+find_astra_registry() {
```
palicena, line 1190: This function and the connector/trident-specific ones below are for determining if we want to create **astra-regcred** or not.

unified-installer/astra-unified-installer.sh
```diff
@@ -1117,12 +1179,177 @@ is_astra_registry() {
...
+find_astra_registry_in_connector_images() {
+    find_astra_registry "$CONNECTOR_IMAGE_REGISTRY" "$NEPTUNE_IMAGE_REGISTRY" "$CONNECTOR_OPERATOR_IMAGE_REGISTRY"
+}
+
+find_astra_registry_in_trident_images() {
+    find_astra_registry "$TRIDENT_OPERATOR_IMAGE_REGISTRY" "$TRIDENT_AUTOSUPPORT_IMAGE_REGISTRY" \
+        "$TRIDENT_IMAGE_REGISTRY" "$TRIDENT_ACP_IMAGE_REGISTRY"
+}
+
+generate_kustomization_dirs_and_files() {
```
palicena, line 1210: This used to be a `step_`, but it's been turned into a generic function since we generate three kustomization files now.

unified-installer/astra-unified-installer.sh
```diff
@@ -1117,12 +1179,177 @@ is_astra_registry() {
...
+
+    if [ "$add_as_resource_to_top_kustomization_file" == "true" ]; then
+        local -r top_kustomization_file="$__GENERATED_KUSTOMIZATION_FILE"
+        local -r base_dir="$(basename "$kustomization_dir")"
+        insert_into_file_after_pattern "$top_kustomization_file" "resources:" "- ./$base_dir"
+        logdebug "$top_kustomization_file: added resource entry for resource at path '$base_dir'"
+    fi
+}
+
+add_astra_registry_secret_to_kustomization() {
```
palicena, line 1238: This was also part of a step before; it was split out from the original step and made generic so that both the connector and trident components can generate the secret in their respective namespaces.

unified-installer/astra-unified-installer.sh
```diff
@@ -1117,12 +1179,177 @@ is_astra_registry() {
...
+          "username": "'"$account_id"'",
+          "password": "'"$api_token"'",
+          "auth": "'"$encoded_creds"'"
+        }
+      }
+    }'
+    logdebug "$kustomization_file: added secret '$pull_secret' to namespace '$namespace'"
+}
+
+add_api_token_secret_to_kustomization() {
```
palicena, line 1275: This is the other part of the step that `add_astra_registry_secret_to_kustomization` was also a part of. This one is only used for the connector, but figured I'd make it a generic function and place it next to the other since they're pretty similar.

unified-installer/astra-unified-installer.sh
```diff
@@ -1117,12 +1179,177 @@ is_astra_registry() {
...
+    # Add API token secret
+    insert_into_file_after_pattern "$kustomization_file" "secretGenerator:" \
+"- name: \"${pull_secret}\"
+  namespace: \"${namespace}\"
+  literals:
+  - apiToken=\"${api_token}\""
+    logdebug "$kustomization_file: added secret '$pull_secret' to namespace '$namespace'"
+}
+
+add_imagepullsecret_patch_to_kustomization() {
```
palicena, line 1296: This is the 3rd part of `step_kustomize_global_pull_secret_if_needed`. This piece is used for both the connector and trident kustomization files, and is for setting the **imagePullSecrets** in the operator deployments.

unified-installer/astra-unified-installer.sh
```diff
@@ -1117,12 +1179,177 @@ is_astra_registry() {
...
+'- target:
+    kind: Deployment
+  patch: |-
+    - op: replace
+      path: /spec/template/spec/imagePullSecrets
+      value:'"${patch_entries}"''
+    logdebug "$kustomization_file: added pull secret patch for [${pull_secrets[*]}]"
+}
+
+add_namespace_transformer_to_kustomization() {
```
palicena, line 1320: This used to be `step_kustomize_global_namespace_if_needed`; as with the others, it was made generic because both the connector and trident kustomization files need it.

unified-installer/astra-unified-installer.sh
```diff
@@ -1708,16 +1935,7 @@ step_check_config() {
     done
 
     # Env vars with special conditions
-    if [ -n "$IMAGE_PULL_SECRET" ]; then
-        if [ -z "$NAMESPACE" ]; then
```
palicena, line 1712: Because `CONNECTOR_NAMESPACE` and `TRIDENT_NAMESPACE` are the source of truth and both have a default, we no longer need this check and prompt since they will never be empty.

unified-installer/astra-unified-installer.sh
```diff
@@ -1857,11 +2077,11 @@ step_check_volumesnapshotclasses() {
 }
 
 step_check_namespace_and_pull_secret_exist() {
-    local -r namespace="$NAMESPACE"
-    local -r pull_secret="$IMAGE_PULL_SECRET"
+    local -r namespace="$1"
+    local -r pull_secret="${2:-""}"
```
palicena, line 2081: Made more generic so that it can be called for both the connector and trident namespace.

unified-installer/astra-unified-installer.sh
```diff
@@ -2104,23 +2328,6 @@ step_check_kubeconfig_choice() {
     fi
 }
 
-# step_query_user_for_resource_count is used to query the user on the number of nodes/namespaces they expect
-# to have on their cluster, which will help us recommend the appropriate resource limit preset.
-step_detect_resource_count() {
```
palicena, line 2109: Dead code.

unified-installer/astra-unified-installer.sh
```diff
@@ -2133,136 +2340,14 @@ step_init_generated_dirs_and_files() {
...
-        cat <<EOF > "$kustomization_file"
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-resources:
-patches:
-transformers:
-EOF
-        logdebug "$kustomization_file: OK"
-    fi
-}
```
palicena, line 2148: The block above has been moved to its own function, `generate_kustomization_dirs_and_files`.

unified-installer/astra-unified-installer.sh
```diff
@@ -2133,136 +2340,14 @@ step_init_generated_dirs_and_files() {
...
-
-step_kustomize_global_namespace_if_needed() {
```
palicena, line 2150: This step got turned into a generic function which is called from the `step_generate_whatever_yaml` function for the connector and trident.

unified-installer/astra-unified-installer.sh
```diff
@@ -2133,136 +2340,14 @@ step_init_generated_dirs_and_files() {
...
-  namespace: "${global_namespace}"
-fieldSpecs:
-- path: metadata/namespace
-  create: true
-EOF
-    insert_into_file_after_pattern "$kustomization_file" "transformers:" "- $transformer_file_name"
-    logdebug "$kustomization_file: added namespace transformer ($transformer_file_name)"
-}
-
-step_kustomize_global_pull_secret_if_needed() {
```
palicena, line 2180: This is the step that got split into 3, with most of those called from the `step_generate_whatever_yaml` functions for the connector and trident.

unified-installer/astra-unified-installer.sh
```diff
@@ -2133,136 +2340,14 @@ step_init_generated_dirs_and_files() {
...
-        - name: "'"${global_pull_secret}"'"
-'
-    logdebug "$kustomization_file: added pull secret patch ($global_pull_secret)"
+    generate_kustomization_dirs_and_files "$kustomization_file"
 }
 
 step_generate_astra_connector_yaml() {
-    local -r kustomization_file="$__GENERATED_KUSTOMIZATION_FILE"
-    local -r kustomization_dir="$(dirname "$kustomization_file")"
+    local -r connector_kustomization_file="$__GENERATED_CONNECTOR_KUSTOMIZATION_FILE"
```
palicena, line 2348: This step no longer operate on the main kustomization file, only the connector kustomization file

unified-installer/astra-unified-installer.sh
```diff
@@ -2488,31 +2592,51 @@ step_existing_trident_flags_compatibility_check() {
 }
 
 step_generate_trident_fresh_install_yaml() {
-    local -r kustomization_file="$__GENERATED_KUSTOMIZATION_FILE"
+    local -r trident_kustomization_file="$__GENERATED_TRIDENT_KUSTOMIZATION_FILE"
```
palicena, line 2595: Same as the connector; this function now only operators on the trident kustomization file.

unified-installer/astra-unified-installer.sh
```diff
@@ -2547,9 +2671,9 @@ step_generate_trident_operator_patch() {
             logdebug "image pull secret '$IMAGE_PULL_SECRET' already present in trident-operator"
         else
             if [ -z "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" ]; then
-                patch_list+=',{"op":"replace","path":"/spec/template/spec/imagePullSecrets","value":[{"name":'"$IMAGE_PULL_SECRET"'}]}'
+                patch_list+=',{"op":"replace","path":"/spec/template/spec/imagePullSecrets","value":[{"name":"'"$IMAGE_PULL_SECRET"'"}]}'
             else
-                patch_list+=',{"op":"add","path":"/spec/template/spec/imagePullSecrets/-","value":{"name":'"$IMAGE_PULL_SECRET"'}}'
+                patch_list+=',{"op":"add","path":"/spec/template/spec/imagePullSecrets/-","value":{"name":"'"$IMAGE_PULL_SECRET"'"}}'
```
palicena, line 2676: This line and the one above were missing literal double quotes`"`, which was breaking the patch.

unified-installer/astra-unified-installer.sh
```diff
@@ -2590,16 +2714,28 @@ step_generate_torc_patch() {
...
+    done
+    if ! is_empty "${pull_secrets_to_add[@]}"; then
+        if [ -z "$_EXISTING_TRIDENT_OPERATOR_PULL_SECRETS" ]; then
+            local -r secrets_list="$(join_str "${pull_secrets_to_add[@]}" ",")"
+            torc_patch_list+='{"op":"replace","path":"/spec/imagePullSecrets","value":['"$secrets_list"']},'
+        else
+            for secret in "${pull_secrets_to_add[@]}" ; do
+                torc_patch_list+='{"op":"add","path":"/spec/imagePullSecrets/-","value":"'"$secret"'"},'
+            done
+        fi
```
palicena, line 2738: The main changes around these parts is determining which (if any) secret out of `IMAGE_PULL_SECRET` and `astra-regcred` we want to include in the list of pull secrets. This involves doing the following for both:
- Check if the secret already exists in the TORC's list of secrets
- If not, add the secret to the list

unified-installer/astra-unified-installer.sh
```diff
@@ -2632,9 +2768,16 @@ step_add_image_remaps_to_kustomization() {
...
+        # Note: depending on which Trident YAML we pulled, the ref sometimes excludes 'docker.io', so
+        # we add remaps both with and without to be safe.
         echo "- name: netapp/$__DEFAULT_TRIDENT_OPERATOR_IMAGE_NAME" >> "$kustomization_file"
         echo "  newName: $(join_rpath "$TRIDENT_OPERATOR_IMAGE_REGISTRY" "$TRIDENT_OPERATOR_IMAGE_REPO")" >> "$kustomization_file"
         echo "  newTag: \"$TRIDENT_OPERATOR_IMAGE_TAG\"" >> "$kustomization_file"
+
+        local -r default_img="$__DEFAULT_DOCKER_HUB_IMAGE_REGISTRY/$__DEFAULT_DOCKER_HUB_IMAGE_BASE_REPO/$__DEFAULT_TRIDENT_OPERATOR_IMAGE_NAME"
+        echo "- name: $default_img" >> "$kustomization_file"
+        echo "  newName: $(join_rpath "$TRIDENT_OPERATOR_IMAGE_REGISTRY" "$TRIDENT_OPERATOR_IMAGE_REPO")" >> "$kustomization_file"
+        echo "  newTag: \"$TRIDENT_OPERATOR_IMAGE_TAG\"" >> "$kustomization_file"
```
palicena, line 2780: When testing, I found that, after a recent change to the Trident YAML we pull, the repository paths have changed from `netapp/trident-operator` to `docker.io/netapp/trident-operator`. Added the new path and leaving the old one in to be safe.

unified-installer/astra-unified-installer.sh
```diff
@@ -2659,41 +2802,50 @@ step_apply_resources() {
...
-                if echo "$line" | grep -q "Warning:"; then
-                    logdebug "captured warning when applying kustomize resources:${__NEWLINE}$captured_err"
-                elif echo "$line" | grep -q "no objects passed to apply"; then
-                    logdebug "no kustomize resources to apply, skipping"
-                elif [ -z "$output" ] || [ -n "$line" ]; then
-                    add_problem "Failed to apply kustomize resources: $line"
-                fi
-            done < <(echo "$captured_err")
+    local -a other_args=()
+    is_dry_run && other_args+=("--dry-run=client")
```
palicena, line 2806: Main change here is replacing the `if ! is_dry_run` check with `--dry-run=client`. This allows us to test the kubectl apply commands in some capacity instead of skipping them during a dry run.

unified-installer/astra-unified-installer.sh
```diff
@@ -2659,41 +2802,50 @@ step_apply_resources() {
...
+
+    # Apply trident-acp secret if it exists
+    if [ -e "$__GENERATED_TRIDENT_ACP_SECRET_FILE" ]; then
+        output="$(kubectl apply -f "$__GENERATED_TRIDENT_ACP_SECRET_FILE" -n "$(get_trident_namespace)" "${other_args[@]}" 2> "$__ERR_FILE")"
+        captured_err="$(get_captured_err | remove_kubectl_warnings_from_error_output)"
```
palicena, line 2811: Other notable change is removing the warnings from the captured errors so we don't get false positives.

unified-installer/astra-unified-installer.sh
```diff
@@ -2659,41 +2802,50 @@ step_apply_resources() {
...
+    else
+        loginfo "* No Astra operator resources to apply, skipping."
+    fi
 
     # Apply CRs (if we have any)
     if [ -f "$crs_file_path" ]; then
         logdebug "apply CRs"
         if ! is_dry_run; then
-            output="$(kubectl apply -f "$crs_file_path" 2> "$__ERR_FILE")"
+            output="$(kubectl apply -f "$crs_file_path" "${other_args[@]}" 2> "$__ERR_FILE")"
```
palicena, line 2848: Same dry run change here.

unified-installer/astra-unified-installer.sh
```diff
@@ -2714,42 +2868,45 @@ step_apply_trident_operator_patches() {
     local -ra patches=("${_PATCHES_TRIDENT_OPERATOR[@]}")
     local -r patches_len="${#patches[@]}"
 
-    if ! trident_will_be_installed_or_modified && [ "$patches_len" -gt 0 ]; then
+    if [ "$patches_len" -eq 0 ]; then
+        loginfo "* No Astra Trident Operator patches to apply, skipping"
+        return 0
```
palicena, line 2873: Exit early if we don't have any patches to apply.

unified-installer/astra-unified-installer.sh
```diff
@@ -2714,42 +2868,45 @@ step_apply_trident_operator_patches() {
...
-    if debug_is_on && [ "$patches_len" -gt 0 ]; then
-        local disclaimer="# THIS FILE IS FOR DEBUGGING PURPOSES ONLY. These are the patches applied to the"
-        disclaimer+="${__NEWLINE}# Astra Trident Operator deployment when upgrading the Astra Trident Operator."
-        disclaimer+="${__NEWLINE}"
-        append_lines_to_file "${__GENERATED_PATCHES_TRIDENT_OPERATOR_FILE}" "$disclaimer" "${patches[@]}"
-    fi
+    local disclaimer="# THIS FILE IS FOR DEBUGGING PURPOSES ONLY. These are the patches applied to the"
+    disclaimer+="${__NEWLINE}# Astra Trident Operator deployment when upgrading the Astra Trident Operator."
+    disclaimer+="${__NEWLINE}"
+    append_lines_to_file "${__GENERATED_PATCHES_TRIDENT_OPERATOR_FILE}" "$disclaimer" "${patches[@]}"
```
palicena, line 2881: For this, I removed the `debug_is_on` requirement for generating the patch files. We already generate every resource we apply, seemed to make sense to generate this as well (even though we apply the patch from an in-memory reference, not a file -- maybe that should change?).

unified-installer/astra-unified-installer.sh
```diff
@@ -2924,7 +3081,7 @@ if trident_will_be_installed_or_modified; then
             if ! acp_is_enabled; then
                 if config_acp_image_is_custom || prompt_user_yes_no "Would you like to enable Astra Control Provisioner?"; then
                     # create trident-acp secret
-                    generate_docker_registry_secret "$IMAGE_PULL_SECRET" "$ASTRA_ACCOUNT_ID" "$ASTRA_API_TOKEN" "$(get_trident_namespace)" "$TRIDENT_ACP_IMAGE_REGISTRY" "$__GENERATED_TRIDENT_ACP_SECRET_FILE"
+                    generate_docker_registry_secret "$__ASTRA_REGISTRY_IMAGE_PULL_SECRET" "$ASTRA_ACCOUNT_ID" "$ASTRA_API_TOKEN" "$(get_trident_namespace)" "$TRIDENT_ACP_IMAGE_REGISTRY" "$__GENERATED_TRIDENT_ACP_SECRET_FILE"
```
palicena, line 3084: Pass in `astra-regcred` instead of `IMAGE_PULL_SECRET` since they are now separate.

